PR #6, opened 2018-01-04
(MODULES-6264)(MODULES-6266) Create V2 API and V1 adapter

glennsarti:
This PR adds a new class and provider which has the same interface as the older V1
Scheduled Task.  This adapter class will implement the V1 methods but using the V2 API calls.

This PR also adds tests to ensure that the adapter behaves the same as the V1 puppet type

This PR also adds some basic integration tests, which may subsume the current acceptance tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,4 @@
-
 # scheduled_task
-
-Welcome to your new module. A short overview of the generated parts can be found in the PDK documentation at https://docs.puppet.com/pdk/1.0/pdk_generating_modules.html#module-contents .
-
-Below you'll find the default README template ready for some content.
-
-
-
-
-
-
 
 #### Table of Contents
 
@@ -25,57 +14,40 @@ Below you'll find the default README template ready for some content.
 
 ## Description
 
-Start with a one- or two-sentence summary of what the module does and/or what problem it solves. This is your 30-second elevator pitch for your module. Consider including OS/Puppet version it works with.       
-
-You can give more descriptive information in a second paragraph. This paragraph should answer the questions: "What does this module *do*?" and "Why would I use it?" If your module has a range of functionality (installation, configuration, management, etc.), this is the time to mention it.
+This module adds a new [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) provider capable of using the more modern Version 2 Windows API for task management.
 
 ## Setup
 
-### What scheduled_task affects **OPTIONAL**
+### Beginning with scheduled_task
 
-If it's obvious what your module touches, you can skip this section. For example, folks can probably figure out that your mysql_instance module affects their MySQL instances.
+The scheduled_task module adapts the Puppet [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) resource to run using a modern API. To get started, install the module and declare 'taskscheduler_api2' as the provider, for example:
 
-If there's more that they should know about, though, this is the place to mention:
-
-* Files, packages, services, or operations that the module will alter, impact, or execute.
-* Dependencies that your module automatically installs.
-* Warnings or other important notices.
-
-### Setup Requirements **OPTIONAL**
-
-If your module requires anything extra before setting up (pluginsync enabled, another module, etc.), mention it here. 
-  
-If your most recent release breaks compatibility or requires particular steps for upgrading, you might want to include an additional "Upgrading" section here.
-
-### Beginning with scheduled_task  
-
-The very basic steps needed for a user to get the module up and running. This can include setup steps, if necessary, or it can be an example of the most basic use of the module.
+~~~ puppet
+scheduled_task { 'Run Notepad':
+  command  => "notepad.exe",
+  ...
+  provider => 'taskscheduler_api2'
+}
+~~~
 
 ## Usage
 
-This section is where you describe how to customize, configure, and do the fancy stuff with your module here. It's especially helpful if you include usage examples and code samples for doing things with your module.
+See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
 
 ## Reference
 
-Users need a complete list of your module's classes, types, defined types providers, facts, and functions, along with the parameters for each. You can provide this list either via Puppet Strings code comments or as a complete list in the README Reference section.
+### Provider
 
-* If you are using Puppet Strings code comments, this Reference section should include Strings information so that your users know how to access your documentation.
+* taskscheduler_api2: Adapts the Puppet scheduled_task resource to use the modern Version 2 API.
 
-* If you are not using Puppet Strings, include a list of all of your classes, defined types, and so on, along with their parameters. Each element in this listing should include:
+### Type
 
-  * The data type, if applicable.
-  * A description of what the element does.
-  * Valid values, if the data type doesn't make it obvious.
-  * Default value, if any.
+* scheduled_task: See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
 
 ## Limitations
 
-This is where you list OS compatibility, version compatibility, etc. If there are Known Issues, you might want to include them under their own heading here.
+* Only supported on Windows Server 2008 and above, and Windows 7 and above.
 
 ## Development
 
-Since your module is awesome, other users will want to play with it. Let them know what the ground rules for contributing are.
-
-## Release Notes/Contributors/Etc. **Optional**
-
-If you aren't using changelog, put your release notes here (though you should consider using changelog). You can also add any additional sections you feel are necessary or important to include here. Please use the `## ` header. 
+Puppet modules on the Puppet Forge are open projects, and community contributions are essential for keeping them great. We can’t access the huge number of platforms and myriad hardware, software, and deployment configurations that Puppet is intended to serve, therefore want to keep it as easy as possible to contribute changes so that our modules work in your environment. There are a few guidelines that we need contributors to follow so that we can have a chance of keeping on top of things. For guidelines on how to contribute, see our [module contribution guide.](https://docs.puppet.com/forge/contributing.html)
--- a/examples/create_task_v1.pp
+++ b/examples/create_task_v1.pp
@@ -1,0 +1,9 @@
+scheduled_task { 'Run Notepad':
+  ensure   => present,
+  command  => 'C:\Windows\System32\notepad.exe',
+  trigger  => {
+    schedule   => daily,
+    start_time => '12:00',
+  },
+  provider => 'taskscheduler_api2'
+}
--- a/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
@@ -1,0 +1,318 @@
+require 'puppet/util/windows'
+
+# The TaskScheduler2 class encapsulates taskscheduler settings and behavior using the v2 API
+# https://msdn.microsoft.com/en-us/library/windows/desktop/aa383600(v=vs.85).aspx
+
+# @api private
+module PuppetX
+module PuppetLabs
+module ScheduledTask
+
+module TaskScheduler2
+  # The name of the root folder for tasks
+  ROOT_FOLDER = '\\'.freeze
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383558(v=vs.85).aspx
+  TASK_ENUM_HIDDEN  = 0x1
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa380596(v=vs.85).aspx
+  TASK_ACTION_EXEC = 0
+  TASK_ACTION_COM_HANDLER = 5
+  TASK_ACTION_SEND_EMAIL = 6
+  TASK_ACTION_SHOW_MESSAGE = 7
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383557(v=vs.85).aspx
+  # Undocumented values
+  # Win7/2008 R2                       = 3
+  # Win8/Server 2012 R2 or Server 2016 = 4
+  # Windows 10                         = 6
+  TASK_COMPATIBILITY_AT = 0
+  TASK_COMPATIBILITY_V1 = 1
+  TASK_COMPATIBILITY_V2 = 2
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383915%28v=vs.85%29.aspx
+  TASK_TRIGGER_EVENT                 = 0
+  TASK_TRIGGER_TIME                  = 1
+  TASK_TRIGGER_DAILY                 = 2
+  TASK_TRIGGER_WEEKLY                = 3
+  TASK_TRIGGER_MONTHLY               = 4
+  TASK_TRIGGER_MONTHLYDOW            = 5
+  TASK_TRIGGER_IDLE                  = 6
+  TASK_TRIGGER_REGISTRATION          = 7
+  TASK_TRIGGER_BOOT                  = 8
+  TASK_TRIGGER_LOGON                 = 9
+  TASK_TRIGGER_SESSION_STATE_CHANGE  = 11
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa382538%28v=vs.85%29.aspx
+  TASK_VALIDATE_ONLY                 = 0x1
+  TASK_CREATE                        = 0x2
+  TASK_UPDATE                        = 0x4
+  TASK_CREATE_OR_UPDATE              = 0x6
+  TASK_DISABLE                       = 0x8
+  TASK_DONT_ADD_PRINCIPAL_ACE        = 0x10
+  TASK_IGNORE_REGISTRATION_TRIGGERS  = 0x20
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383566(v=vs.85).aspx
+  TASK_LOGON_NONE                           = 0
+  TASK_LOGON_PASSWORD                       = 1
+  TASK_LOGON_S4U                            = 2
+  TASK_LOGON_INTERACTIVE_TOKEN              = 3
+  TASK_LOGON_GROUP                          = 4
+  TASK_LOGON_SERVICE_ACCOUNT                = 5
+  TASK_LOGON_INTERACTIVE_TOKEN_OR_PASSWORD  = 6
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa380747(v=vs.85).aspx
+  TASK_RUNLEVEL_LUA     = 0
+  TASK_RUNLEVEL_HIGHEST = 1
+
+  RESERVED_FOR_FUTURE_USE = 0
+
+  def self.folder_path_from_task_path(task_path)
+    path = task_path.rpartition('\\')[0]
+
+    path.empty? ? ROOT_FOLDER : path
+  end
+
+  def self.task_name_from_task_path(task_path)
+    task_path.rpartition('\\')[2]
+  end
+
+  # Returns an array of scheduled task names.
+  # By default EVERYTHING is enumerated
+  # option hash
+  #    include_child_folders: recurses into child folders for tasks. Default true
+  #    include_compatibility: Only include tasks which have any of the specified compatibility levels. Default empty array (everything is permitted)
+  #
+  def self.enum_task_names(folder_path = ROOT_FOLDER, enum_options = {})
+    raise TypeError unless folder_path.is_a?(String)
+
+    options = {
+      :include_child_folders => true,
+      :include_compatibility => [],
+    }.merge(enum_options)
+
+    array = []
+
+    task_folder = task_service.GetFolder(folder_path)
+    filter_compatibility = !options[:include_compatibility].empty?
+    task_folder.GetTasks(TASK_ENUM_HIDDEN).each do |task|
+      next if filter_compatibility && !options[:include_compatibility].include?(task.Definition.Settings.Compatibility)
+      array << task.Path
+    end
+    return array unless options[:include_child_folders]
+
+    task_folder.GetFolders(RESERVED_FOR_FUTURE_USE).each do |child_folder|
+      array += enum_task_names(child_folder.Path, options)
+    end
+
+    array
+  end
+
+  def self.task(task_path)
+    raise TypeError unless task_path.is_a?(String)
+
+    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+
+    task_object = task_folder.GetTask(task_name_from_task_path(task_path))
+
+    task_object
+  end
+
+  def self.new_task_definition
+    task_service.NewTask(0)
+  end
+
+  def self.task_definition(task)
+    definition = task_service.NewTask(0)
+    definition.XmlText = task.XML
+
+    definition
+  end
+
+  # Creates or Updates an existing task with the supplied task definition
+  # If task_object is a string then this is a new task and the supplied object is the new task's full path
+  # Otherwise we expect a Win32OLE Task object to be passed through
+  def self.save(task_object, definition, password = nil)
+    task_path = task_object.is_a?(String) ? task_object : task_object.Path
+
+    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+    task_user = nil
+    task_password = nil
+
+    case definition.Principal.LogonType
+      when TASK_LOGON_PASSWORD, TASK_LOGON_INTERACTIVE_TOKEN_OR_PASSWORD
+        task_user = definition.Principal.UserId
+        task_password = password
+    end
+    task_folder.RegisterTaskDefinition(task_name_from_task_path(task_path),
+                                       definition, TASK_CREATE_OR_UPDATE, task_user, task_password,
+                                       definition.Principal.LogonType)
+  end
+
+  # Delete the specified task name.
+  #
+  def self.delete(task_path)
+    raise TypeError unless task_path.is_a?(String)
+    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+
+    task_folder.DeleteTask(task_name_from_task_path(task_path),0)
+  end
+
+  # General Properties
+  def self.principal(definition)
+    definition.Principal
+  end
+
+  def self.set_principal(definition, user)
+    if (user.nil? || user == "")
+      # Setup for the local system account
+      definition.Principal.UserId = 'SYSTEM'
+      definition.Principal.LogonType = TASK_LOGON_SERVICE_ACCOUNT
+      definition.Principal.RunLevel = TASK_RUNLEVEL_HIGHEST
+      return true
+    else
+      definition.Principal.UserId = user
+      definition.Principal.LogonType = TASK_LOGON_PASSWORD
+      definition.Principal.RunLevel = TASK_RUNLEVEL_HIGHEST
+      return true
+    end
+  end
+
+  # Returns the compatibility level of the task.
+  #
+  def self.compatibility(definition)
+    definition.Settings.Compatibility
+  end
+
+  # Sets the compatibility with the task.
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa381846(v=vs.85).aspx
+  #
+  def self.set_compatibility(definition, value)
+    definition.Settings.Compatibility = value
+  end
+
+  # Task Actions
+  # Returns the number of actions associated with the active task.
+  #
+  def self.action_count(definition)
+    definition.Actions.count
+  end
+
+  def self.action(definition, index)
+    result = nil
+
+    begin
+      result = definition.Actions.Item(index)
+    rescue WIN32OLERuntimeError => err
+      # E_INVALIDARG 0x80070057 from # https://msdn.microsoft.com/en-us/library/windows/desktop/aa378137%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
+      if err.message =~ /80070057/m
+        result = nil
+      else
+        raise
+      end
+    end
+
+    result
+  end
+
+  def self.create_action(definition, action_type)
+    definition.Actions.Create(action_type)
+  end
+
+  # Task Triggers
+  def self.trigger_count(definition)
+    definition.Triggers.count
+  end
+
+  # Returns a Win32OLE Trigger Object for the trigger at the given index for the
+  # supplied definition.
+  #
+  # Returns nil if the index does not exist
+  #
+  # Note - This is a 1 based array (not zero)
+  #
+  def self.trigger(definition, index)
+    result = nil
+
+    begin
+      result = definition.Triggers.Item(index)
+    rescue WIN32OLERuntimeError => err
+      # E_INVALIDARG 0x80070057 from # https://msdn.microsoft.com/en-us/library/windows/desktop/aa378137%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
+      if err.message =~ /80070057/m
+        result = nil
+      else
+        raise
+      end
+    end
+
+    result
+  end
+
+  def self.append_new_trigger(definition, trigger_type)
+    definition.Triggers.create(trigger_type)
+  end
+
+  # Deletes the trigger at the specified index.
+  #
+  def self.delete_trigger(definition, index)
+    definition.Triggers.Remove(index)
+
+    index
+  end
+
+  # Helpers
+
+  # From https://msdn.microsoft.com/en-us/library/windows/desktop/aa381850(v=vs.85).aspx
+  # https://en.wikipedia.org/wiki/ISO_8601#Durations
+  #
+  # The format for this string is PnYnMnDTnHnMnS, where nY is the number of years, nM is the number of months,
+  # nD is the number of days, 'T' is the date/time separator, nH is the number of hours, nM is the number of minutes,
+  # and nS is the number of seconds (for example, PT5M specifies 5 minutes and P1M4DT2H5M specifies one month,
+  # four days, two hours, and five minutes)
+  def self.duration_to_hash(duration)
+    regex = /^P((?'year'\d+)Y)?((?'month'\d+)M)?((?'day'\d+)D)?(T((?'hour'\d+)H)?((?'minute'\d+)M)?((?'second'\d+)S)?)?$/
+
+    matches = regex.match(duration)
+    return nil if matches.nil?
+
+    {
+      :year => matches['year'],
+      :month => matches['month'],
+      :day => matches['day'],
+      :minute => matches['minute'],
+      :hour => matches['hour'],
+      :second => matches['second'],
+    }
+  end
+
+  # Converts a hash table describing year, month, day etc. into a timelimit string as per
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa381850(v=vs.85).aspx
+  # https://en.wikipedia.org/wiki/ISO_8601#Durations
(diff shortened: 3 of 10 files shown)


glennsarti:
Can't repro the failures locally so will need to break into an appveyor build insitu.

Iristyle:
This is the artist formerly known as [PUP-7894](https://tickets.puppetlabs.com/browse/PUP-7894) - originally proposed in PR at https://github.com/puppetlabs/puppet/pull/6382

glennsarti:
@clairecadman When you get a moment, can you give the README a once-over.  It's a completely new README so feel free to comment on anything in it, not just my changes.

clairecadman:
@glennsarti Added a couple of comments.

glennsarti:
@clairecadman I've updated the README.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,279 @@
+# This class is used to manage V1 compatible tasks using the Task Scheduler V2 API
```
Iristyle: Nit - this is not a Facade, it's an Adapter.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,596 @@
...
+  #
+  def delete(task_name)
+    full_taskname = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER + normalize_task_name(task_name)
+    @tasksched.delete(full_taskname)
+  end
+
+  # Execute the current task.
+  #
+  def run
+    raise NotImplementedError
```
Iristyle: All of these `raise NotImplementedError` can be deleted as they're dead code without any callers. Don't need to handle in this PR, can defer to next cut...

glennsarti: Yeah... this was a tricky one in my head.  Do I only implement the minimum methods or make this object a direct replacement for the V1 object.  I landed with this so at least if anyone was using those methods, they'd throw Not Implemented as opposed to Method Missing.

But you're right this is moot.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: I'll reword.  Aren't metaphors wonderful!  People can have different interpretations.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
Iristyle: Yeah, I know we talked about this before, and then the plan changed a bit along the way. The thing is, to use these methods users would have to know they can swap this adapter in for the old class, which no one can possibly be aware of yet ;0

YAGNI and all that...

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: Reworded.

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -109,7 +111,20 @@ def time_component
...
-describe Puppet::Type.type(:scheduled_task).provider(:win32_taskscheduler), :if => Puppet.features.microsoft_windows? do
+# The Win32::TaskScheduler and PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task classes should be
+# API compatible and behave the same way.  What differs is which Windows API is used to query
+# and affect the system.  This means for testing, any tests should be the same no matter what
+# provider or concrete class (which the provider uses) is used.
+klass_list = Puppet.features.microsoft_windows? ? [Win32::TaskScheduler, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task] : []
+klass_list.each do |concrete_klass|
+
+if concrete_klass == PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task
+  task_provider = :taskscheduler_api2
```
Iristyle: The one thing I'm not totally keen on is that the constants used in this class are still scoped to `Win32::TaskScheduler::` rather than `concrete_klass::`, so we have a bunch of leakage from v1 provider into v2 provider.

Given what this PR is for now, should be OK for the moment... but we should probably bookmark this as something to loop back on / clean up later. I understand why you want the constants to be the same values for the sake of API compat... 

FWIW, the names of the providers as old `Win32::TaskScheduler` and new `PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task` don't make it clear that they're related to one another either, but not sure that's something we can or want to change really.

glennsarti: I can add a TODO: comment on those

We certainly can't change the 'Win32::TaskScheduler' and that's a hangover of removing the win32-taskscheduler gem.  I strongly feel that changing the name `PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task` is definitely where it should be, rather than say 'Win32::TaskScheduler2' or something

Iristyle: I think what we would want to do is move / rename constants into the new world, and provide aliases from old world to new world for backwards compat, with some timeline on removing the old things (i.e. Puppet 6). It's unlikely that anyone outside  of Pupet is using the old `Win32::TaskScheduler` namespace (we can scan Forge to be absolutely certain though).

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+# https://msdn.microsoft.com/en-us/library/windows/desktop/aa383600(v=vs.85).aspx
+
+# @api private
+module PuppetX
+module PuppetLabs
+module ScheduledTask
+
+module TaskScheduler2
+  # The error class raised if any task scheduler specific calls fail.
+  # class Error < Puppet::Util::Windows::Error; end
```
Iristyle: Whoops, we've got a clinger!

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+    raise TypeError unless folder_path.is_a?(String)
+
+    options[:include_child_folders] = true if options[:include_child_folders].nil?
+    options[:include_compatibility] = [] if options[:include_compatibility].nil?
+
+    array = []
+
+    task_folder = task_service.GetFolder(folder_path)
+    task_folder.GetTasks(TASK_ENUM_HIDDEN).each do |task|
+      included = true
```
Iristyle: Would be nice to simplify how this reads... how about something like?

```ruby
filter_compatibility = !options[:include_compatibility].empty?
task_folder.GetTasks(TASK_ENUM_HIDDEN).each do |task|
  next if filter_compatibility && !options[:include_compatibility].include?(task.Definition.Settings.Compatibility)
  array << task.Path
end
```

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+    duration += hash[:minute].to_s + 'M' unless hash[:minute].nil? || hash[:minute].zero?
+    duration += hash[:second].to_s + 'S' unless hash[:second].nil? || hash[:second].zero?
+
+    duration == 'PT' ? 'PT0S' : duration
+  end
+
+  # Private methods
+  def self.task_service
+    if @@service_object.nil?
+      @@service_object = WIN32OLE.new('Schedule.Service')
```
Iristyle: Will an effective singleton here cause us any problems? Are these expensive to create?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+    task_folder = task_service.GetFolder(folder_path)
+    task_folder.GetTasks(TASK_ENUM_HIDDEN).each do |task|
+      included = true
+      included = included && options[:include_compatibility].include?(task.Definition.Settings.Compatibility) unless options[:include_compatibility].empty?
+
+      array << task.Path if included
+    end
+    return array unless options[:include_child_folders]
+
+    task_folder.GetFolders(0).each do |child_folder|
```
Iristyle: I had a tough time figuring out what the 0 meant... probably because this is a terrible API.

Could we at least name the `0` as `reserved_flag` or add a comment or something?
https://msdn.microsoft.com/en-us/library/windows/desktop/aa381350(v=vs.85).aspx

Otherwise this looks like some sort of count of folders to return or something, which it definitely isn't!

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+    raise TypeError unless task_path.is_a?(String)
+
+    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+
+    task_object = task_folder.GetTask(task_name_from_task_path(task_path))
+
+    task_object
+  end
+
+  def self.definition(task_object = nil)
```
Iristyle: Is it too late to name this something else?

I'm concerned that this method is named `definition`, but that we also have parameters passed to a number of functions that are also named `definition`. Based on scoping rules, we pick up the parameter, but it would be pretty easy to accidentally mess this up in the future.  I think no collision would be preferable.

Iristyle: Similarly naming a variable inside a method the same name as the method is usually not preferred for the same reasons (sometimes leads to stack overflows).

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+    definition = task_service.NewTask(0)
```
Iristyle: Another weird reserved flag?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+
+    result == Puppet::Util::Windows::COM::S_OK
+  end
+
+  # General Properties
+  def self.principal(definition)
+    definition.Principal
+  end
+
+  def self.set_principal(definition, user)
```
Iristyle: Why `set_principal` rather than `principal=`?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
Iristyle: One more comment on this method... I see it's basically designed to give you either a new definition or to clone an existing definition if you pass a task object.  Wondering if it might be better to separate the two into something like `empty_definition` and `task_definition(task)` or something like that?

I understand what you were going for... could probably go either way.

Is only existing usage in specs?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+                                       definition.Principal.LogonType)
+  end
+
+  # Delete the specified task name.
+  #
+  def self.delete(task_path)
+    raise TypeError unless task_path.is_a?(String)
+    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+
+    result = task_folder.DeleteTask(task_name_from_task_path(task_path),0)
```
Iristyle: COM wrappers automatically raised for HRESULT < 0 failures on method calls like this (like https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/taskscheduler.rb#L265) with a `Puppet::Util::Windows::Error`.
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/com.rb#L141-L143

I *think* we should probably continue to do the same here?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -0,0 +1,318 @@
...
+
+  # From https://msdn.microsoft.com/en-us/library/windows/desktop/aa381850(v=vs.85).aspx
+  # https://en.wikipedia.org/wiki/ISO_8601#Durations
+  #
+  # The format for this string is PnYnMnDTnHnMnS, where nY is the number of years, nM is the number of months,
+  # nD is the number of days, 'T' is the date/time separator, nH is the number of hours, nM is the number of minutes,
+  # and nS is the number of seconds (for example, PT5M specifies 5 minutes and P1M4DT2H5M specifies one month,
+  # four days, two hours, and five minutes)
+  def self.duration_to_hash(duration)
+    regex = /^P((?'year'\d+)Y)?((?'month'\d+)M)?((?'day'\d+)D)?(T((?'hour'\d+)H)?((?'minute'\d+)M)?((?'second'\d+)S)?)?$/
```
Iristyle: So I suspect we don't have to worry about this because we always get back the format that we want, with everything specified... but... it does look like something like `PT36H` is valid... which will cause some of the matches to be `nil`

```
[5] pry(main)> regex = /^P((?'year'\d+)Y)?((?'month'\d+)M)?((?'day'\d+)D)?(T((?'hour'\d+)H)?((?'minute'\d+)M)?((?'second'\d+)S)?)?$/
=> /^P((?'year'\d+)Y)?((?'month'\d+)M)?((?'day'\d+)D)?(T((?'hour'\d+)H)?((?'minute'\d+)M)?((?'second'\d+)S)?)?$/
[6] pry(main)> matches = regex.match('PT36H')
=> #<MatchData "PT36H" year:nil month:nil day:nil hour:"36" minute:nil second:nil>
[7] pry(main)> matches['month']
=> nil
```

I didn't see any problems around this when testing so I *think* we're OK?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,596 @@
...
+    end
+  end
+  alias :tasks :enum
+
+  # Activate the specified task.
+  #
+  def activate(task_name)
+    raise TypeError unless task_name.is_a?(String)
+    normal_task_name = normalize_task_name(task_name)
+    raise Error.new(_("Scheduled Task %{task_name} does not exist") % { task_name: normal_task_name }) unless exists?(normal_task_name)
```
Iristyle: Do we have a test for this bit?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,596 @@
...
+  def default_action(create_if_missing = false)
+    action = nil
+    action_index = 1
+    while action_index <= @tasksched.action_count(@definition) do
+      action = @tasksched.action(@definition, action_index)
+      break if action.Type == PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_ACTION_EXEC
+      action = nil
+      action_index += 1
+    end
+    return action unless action.nil?
```
Iristyle:
```ruby
if action.nil? && create_if_missing
  action = @tasksched.create_action(@definition, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_ACTION_EXEC)  
end

action
```

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,596 @@
...
+    task_name = task_name.slice(0,task_name.length - 4) if task_name.end_with?('.job')
+
+    task_name
+  end
+
+  # Find the first TASK_ACTION_EXEC action
+  def default_action(create_if_missing = false)
+    action = nil
+    action_index = 1
+    while action_index <= @tasksched.action_count(@definition) do
```
Iristyle: A little verbose, but a little better... 

```ruby
action = nil

(1..@tasksched.action_count(@definition)) do |i|
  index_action = @tasksched.action(@definition, i)
  action = index_action if index_action.Type == PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_ACTION_EXEC
  break if action
end
```

But I think we can do one better by using `lazy` here to only create items in the map when we don't get matches - here's the entire method (verified locally but only with an `action_count` of 1):

```ruby
def default_action(create_if_missing = false)
  action = (1..@tasksched.action_count(@definition)).lazy.map do |i|
    @tasksched.action(@definition, i)
  end.find do |a|
    a.Type == PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_ACTION_EXEC
  end

  if action.nil? && create_if_missing
    action = @tasksched.create_action(@definition, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_ACTION_EXEC)
  end

  action
end
```

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,596 @@
...
+  def normalize_task_name(task_name)
+    # The Puppet provider and some other instances may pass a '.job' suffix as per the V1 API
+    # This is not needed for the V2 API so we just remove it
+    task_name = task_name.slice(0,task_name.length - 4) if task_name.end_with?('.job')
+
+    task_name
+  end
+
+  # Find the first TASK_ACTION_EXEC action
+  def default_action(create_if_missing = false)
```
Iristyle:
```ruby
def default_action(create_if_missing = false)
```

Would be clearer at callsites as

```ruby
def default_action(create_if_missing: false)
```

Then callers use

```ruby
foo = default_action(create_if_missing: true)
```

examples/create_task_v1.pp
```diff
@@ -0,0 +1,9 @@
+scheduled_task { 'Run Notepad':
+  ensure   => present,
+  command  => 'C:\Windows\System32\notepad.exe',
```
Iristyle: Just what every system should do at noon every day... launch a notepad!

README.md
```diff
@@ -25,57 +14,40 @@ Below you'll find the default README template ready for some content.
...
-This section is where you describe how to customize, configure, and do the fancy stuff with your module here. It's especially helpful if you include usage examples and code samples for doing things with your module.
+See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
 
 ## Reference
 
-Users need a complete list of your module's classes, types, defined types providers, facts, and functions, along with the parameters for each. You can provide this list either via Puppet Strings code comments or as a complete list in the README Reference section.
+### Provider
 
-* If you are using Puppet Strings code comments, this Reference section should include Strings information so that your users know how to access your documentation.
+* taskscheduler_api2: Adapts the Puppet scheduled_task resource to use the modern Version 2 API.
```
Iristyle: Type has backticks, but this provider does not...

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+  task_name
+end
+
+describe "PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2", :if => Puppet.features.microsoft_windows? do
+  let(:subject_taskname) { nil }
+  let(:subject) { PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2 }
+
+  describe '#enum_task_names' do
+    before(:all) do
```
Iristyle: Do you want to do `:all` or `:each` here?

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+    it 'should only return compatible tasks if specified' do
+      subject_count = subject.enum_task_names(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER, { :include_compatibility => [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V1]}).count
+      ps_cmd = '(Get-ScheduledTask | ? { [Int]$_.Settings.Compatibility -eq 1 } | Measure-Object).count'
+      expect(subject_count).to be_same_as_powershell_command(ps_cmd)
+    end
+  end
+
+  describe '#delete' do
+    before(:all) do
```
Iristyle: I would probably change this from `:all` to `:each` in the event more tests are added to this `describe`

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+      end
+
+      it 'should have the specified action arguments' do
+        expect(subject.action(task_definition, 1).Arguments).to eq('/c exit 0')
+      end
+    end
+  end
+
+  describe 'modify a task' do
+    before(:all) do
```
Iristyle: Same here for `:each` instead of `:all` ... if more tests are added, you end up sharing the task which is probably undesirable.

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+# These integration tests confirm that the tasks created in a V1 scheduled task APi are visible
+# in the V2 API, and that changes in the V2 API will appear in the V1 API.
+
+describe "PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task", :if => Puppet.features.microsoft_windows? do
+  let(:subjectv1) { Win32::TaskScheduler.new() }
+  let(:subjectv2) { PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new() }
+
+  let(:command) { 'cmd.exe' }
+  let(:arguments_before) { '/c exit 0' }
```
Iristyle: Never used

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      obj.delete(@task_name) if obj.exists?(@task_name)
+    end
+
+    it 'should be visible by the V2 API' do
+      expect(subjectv2.exists?(@task_name)).to be true
+    end
+
+    it 'should have same properties in the V1 API' do
+      subjectv2.activate(@task_name)
+      subjectv2.parameters = arguments_after
```
Iristyle: `arguments_after` only used here.. probably don't need to be in the `let` above

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Clinger?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Done

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Given this is in a module I'm not sure why I used `@@`.  Should probably just use `@`

Creating OLE objects is "expensive" but relative to a puppet run, probably not.  It just seemed a waste to continuously create and destroy that object, and just keep it around.

glennsarti: For now I've changed it to `@`

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Created a const called `RESERVED_FOR_FUTURE_USE`

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Agreed.  Changed method name to `task_definition`

glennsarti: Surprised rubocop didn't complain....

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Whoops. Agreed.  I broke the (S)olid :-) .  Changed to `new_task_definition` and `task_definition(task)`

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Yeah...as these are WIN32OLERuntimeError not COM errors things are not as clear cut.  The WIN32OLERuntimeError stuff is horrible to work with.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: As I'm converting this to a hash, the missing `months` would be a nil anyway e.g.

` xxxxx.duration_to_hash('PT36H')[:months]
=> nil
`

So either way, they'll be nil.

Currently the V1 API uses milliseconds not the duration format so the tests are safe.  Perhaps when we use the V2 API things may be tripped up e.g.

'PT2M' == 'PT120S'  (2 minutes equals 120 seconds)

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: Apparently not.

glennsarti: Looks like it's already guarded in the new provider
lib/puppet/provider/scheduled_task/taskscheduler_api2.rb:32
`@task.activate(resource[:name] + '.job') if exists?`

and old provider
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/scheduled_task/win32_taskscheduler.rb#L34

glennsarti: Do we need one?

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: Fixed

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: I like the first one.  The lazy code looks harder to grok.

README.md
glennsarti: Fixed

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_spec.rb
glennsarti: Done

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_spec.rb
glennsarti: Done

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task_spec.rb
glennsarti: Oh yeah...that's because I moved them to before(:all) and `let` variables are not available there.

glennsarti: Fixed

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task_spec.rb
glennsarti: Fixed

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
Iristyle: Maybe you started with `@@` when this code was a class / before changing to module?

I was mostly just concerned with understanding the appropriate usage pattern for that COM component.... For instance, I was wondering if it's possible to get the object into a state where a failure in one resource ends up putting that object into a state that makes it unable to service other resources in the same run, etc. I would hope that would not be the case, but given the v1 weirdness, thought it was worth asking the question.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Oh. commented out code.

glennsarti:
@Iristyle Comments addressed.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
Iristyle: I don't think we have it turned on?  I was actually going to suggest enabling it on this project as a starter to see what happens...

I don't see it in the `release_checks` task at https://github.com/puppetlabs/puppetlabs_spec_helper/blob/2403747514bc1aecddccde203e79680f8067b9eb/lib/puppetlabs_spec_helper/rake_tasks.rb#L614-L624

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: In theory yes. It could not recover if the object was in a bad state, whereas if I created and connected on every invocation then it would auto-correct

glennsarti: However that service object shouldn't really store state between calls, it's all about;
1. call `get` task
2. modify the object returned using its methods etc.
3. call `set` task

I'm happy to change it to create a new object on each call.  I suspect the actual time difference is minimal as specs use the code in a different way than a puppet resource.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
Iristyle: Yeah, I now recall our discussion about the `WIN32OLERuntimeError` business and the lack of error codes as an attribute of the error instance... so that's a problem for sure.

But in particular I was looking at the ` == Puppet::Util::Windows::COM::S_OK`, which I assumed indicated you got back a `HRESULT`... However, looking at the docs now, I don't think that that's right - `DeleteTask` says `this method does not return a value`, so I think if nothing is raised you just `return true` at the end:

https://msdn.microsoft.com/en-us/library/windows/desktop/aa382566%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

This is confirmed by the COM interface docs as well https://msdn.microsoft.com/en-us/library/windows/desktop/aa381345(v=vs.85).aspx

Note: All COM interface methods return a `HRESULT` status code by convention, and if they have any return values from the method they're actually `[out]` parameters to pointer-pointers in the IDL.  I assume `WIN32OLERuntimeError` probably already does the `FAILED` check under the hood for us, so my original point is moot I think.

Iristyle: Additional note: I am now a little concerned that this method is always returning `false`, yet we have no test failures to show for it.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
Iristyle: Probably not.. I just noticed that the logic here was a little different that previous helper, especially wrt normalization.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -0,0 +1,595 @@
...
+  def account_information
+    principal = @tasksched.principal(@definition)
+    principal.nil? ? nil : principal.UserId
+  end
+
+  # Returns the name of the application associated with the task.
+  #
+  def application_name
+    action = default_action(create_if_missing: false)
+    action.nil? ? nil : action.Path
```
Iristyle: It will be nice when we can use the safe navigation operator!

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
Iristyle: The `lazy` can be removed in the above btw -- that was an optimization specifically to not fill the map prior to finding the first matching item (which is closer to the behavior of the prior example).

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
glennsarti: Weird.  competing docs -
https://msdn.microsoft.com/en-us/library/windows/desktop/aa381345(v=vs.85).aspx

```
Return value
If this method succeeds, it returns S_OK. Otherwise, it returns an HRESULT error code.
```

glennsarti: Modified delete method
```
  def self.delete(task_path)
    raise TypeError unless task_path.is_a?(String)
    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))

    task_folder.DeleteTask(task_name_from_task_path(task_path),0)
  end
```

and tests
```
    it 'should delete a task that exists' do
...
    it 'should raise an error for a task that does not exist' do
...
```

Iristyle: Just for the sake of anyone else following the thread, the docs are not competing, just  targeted at different consumers.  COM methods all returns status HRESULTs as mentioned in https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/6#discussion_r160872843

For COM functions to return, for instance, new objects, they have additional `[out]` parameters that are passed into the function as pointer pointers... For instance,

```cpp
HRESULT GetFolder(
  [in]  BSTR        path,
  [out] ITaskFolder **ppFolder
);
```

Essentially you pass the address of your `ITaskFolder *` (interface pointer), so that the calling function can write the actual pointer to the memory address you've given when it creates the output object.

Because throwing C++ exceptions out of COM functions is not supported (what would catching  a C++ exception in Visual Basic even mean?), status is conveyed by the convention of always returning a `HRESULT` value for success / fail.

When we use `WIN32OLE` in Ruby, we're using the scripting compatible interfaces, which support late-binding and going through the `IDispatch` interface... so when a function has no `[out]` params in the actual IDL, that means it's basically a `void`.  But the easiest thing to do is just look at the "scripting" docs at https://msdn.microsoft.com/en-us/library/windows/desktop/aa383607(v=vs.85).aspx rather than the C++ docs.

PR #7, opened 2018-01-05
(MODULES-6362) Add acceptance tests

glennsarti:
Previously the acceptance tests were for vanilla beaker.  As this is now a
module, we use beaker-rspec.  This commit converts the create and destroy tests
to beaker-rspec and removes the query and modify tests as they are not required.

Builds on PR #6


README.md
```diff
@@ -25,57 +14,40 @@ Below you'll find the default README template ready for some content.
...
 
 This section is where you describe how to customize, configure, and do the fancy stuff with your module here. It's especially helpful if you include usage examples and code samples for doing things with your module.
 
 ## Reference
 
-Users need a complete list of your module's classes, types, defined types providers, facts, and functions, along with the parameters for each. You can provide this list either via Puppet Strings code comments or as a complete list in the README Reference section.
+### Provider
 
-* If you are using Puppet Strings code comments, this Reference section should include Strings information so that your users know how to access your documentation.
+* taskscheduler_api2: Adapts the Puppet scheduled_task resource to use the modern V2 API.
```
clairecadman: Maybe keep consistent with either saying 'modern' API or 'newer' API, if these are the same API? 

Same with 'V2' and 'Version 2'.

README.md
```diff
@@ -25,57 +14,40 @@ Below you'll find the default README template ready for some content.
...
+* Only supported on Windows Server 2008 and above, and Windows 7 and above.
 
 ## Development
 
-Since your module is awesome, other users will want to play with it. Let them know what the ground rules for contributing are.
-
-## Release Notes/Contributors/Etc. **Optional**
-
-If you aren't using changelog, put your release notes here (though you should consider using changelog). You can also add any additional sections you feel are necessary or important to include here. Please use the `## ` header. 
+Puppet modules on the Puppet Forge are open projects, and community contributions are essential for keeping them great. We can’t access the huge number of platforms and myriad hardware, software, and deployment configurations that Puppet is intended to serve. We want to keep it as easy as possible to contribute changes so that our modules work in your environment. There are a few guidelines that we need contributors to follow so that we can have a chance of keeping on top of things. For more information, see our [module contribution guide.](https://docs.puppet.com/forge/contributing.html)
```
clairecadman: Suggested rewording:

"Puppet modules on the Forge are open projects, and community contributions are essential for their success".

Slight rewording, which is how I read this, if this is what you meant: 

"We can’t access the huge number of platforms and myriad hardware, software, and deployment configurations that Puppet is intended to serve, therefore want to keep it as easy as possible to contribute changes so that our modules work in your environment." 

Suggestion for the last sentence: 

"For guidelines on how to contribute, see our [module contribution guide.](https://docs.puppet.com/forge/contributing.html)"

spec/acceptance/should_create_task_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      arguments   => "foo bar baz",
+      working_dir => 'c:\\\\windows',
+      trigger => {
+        schedule   => daily,
+        start_time => '12:00',
+      },
+      provider    => 'taskscheduler_api2'
+    }
+    MANIFEST
+    execute_manifest(pp, :catch_failures => true)
```
Iristyle: Can we use `execute_manifest_on` to be more explicit about which nodes are actually executing the manifest?

IMHO, my expectation for that API is that *all* Beaker nodes under test will execute the manifest.  Unfortunately, that's not the case. This may be obvious to people familiar with beaker-testmodeswitcher, but I'd argue it's misleading because it only uses the `default` as @RandomNoun7 and I discovered while reviewing this.  I think this can be a problem when the intent is to execute across a number of hosts (i.e. different Windows versions), and would prefer we remove the implicit `default` for clarity.  Obviously this isn't a problem for masterless when there is only one node, but for all other scenarios (multiple agents, master + agent), etc I think it is.

https://github.com/puppetlabs/beaker-testmode_switcher/blob/c789e6fe7203d19674f5f2181a0d1d8e637cdf11/lib/beaker/testmode_switcher/beaker_runners.rb#L74-L76

glennsarti: Updated to `execute_manifest_on`

glennsarti:
@Iristyle updated

PR #10, opened 2018-01-12
(MODULES-6362) Update tests for Server 2016 Core

glennsarti:
Windows Server 2016 core outputs a different error message from schtask.exe.
This caused the acceptance tests to file as the regex mathcers were too
specific.  This commit modifies the regex to be more leniant on the error
message checking.


glennsarti:
Running through adhoc...

PR #12, opened 2018-01-12
Fix metadata which points to iis

TraGicCode:
Currently the puppetforge has the 'Project URL' link which is pointing to the puppetlabs-iis module.  This fixes that :D.


Iristyle:
Thanks!

PR #13, opened 2018-02-02
(MODULES-6843) Refactor Trigger code from V1 -> V2 adapter class

Iristyle:
#14 should be merged first which restores existing type / provider to this module

- Report on the compatibility level of tasks enumerated from the
   system.

   This requires adding a value to the scheduled_task type named
   compatibility. Based on how the provider works, it currently only
   enumerates v1 tasks, and therefore only returns 1.

   Since the property may now also be set in a manifest, limit the
   value to 1 so that it cannot be set improperly. Default the value
   to 1 to preserve backward compatibility.

   `puppet resource` will now emit the value as well

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
@@ -263,14 +263,14 @@ class TaskScheduler2V1Task
     case v1trigger['trigger_type']
       when :TASK_TIME_TRIGGER_DAILY
         # https://msdn.microsoft.com/en-us/library/windows/desktop/aa446858(v=vs.85).aspx
-        trigger_object = @tasksched.append_new_trigger(@definition, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_TRIGGER_DAILY)
+        trigger_object = @definition.Triggers.Create(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_TRIGGER_DAILY)
         trigger_object.DaysInterval = trigger_settings['days_interval']
         # Static V2 settings which are not set by the Puppet scheduledtask type
         trigger_object.Randomdelay = 0
 
       when :TASK_TIME_TRIGGER_WEEKLY
         # https://msdn.microsoft.com/en-us/library/windows/desktop/aa384019(v=vs.85).aspx
-        trigger_object = @tasksched.append_new_trigger(@definition, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_TRIGGER_WEEKLY)
+        trigger_object = @definition.Triggers.Create(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_TRIGGER_WEEKLY)
         trigger_object.DaysOfWeek = trigger_settings['days_of_week']
         trigger_object.WeeksInterval = trigger_settings['weeks_interval']
         # Static V2 settings which are not set by the Puppet scheduledtask type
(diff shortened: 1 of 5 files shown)


Iristyle:
On 5f952d2 it looks like previous history didn't get restored properly for the file. Might have to try reverting the deletion commit instead. Hmmph....

spec/unit/puppet/type/scheduled_task_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+  end
+
+  describe 'when setting the compatibility' do
+    it 'should allow 1' do
+      expect(described_class.new(
+        :title        => 'Foo',
+        :command      => 'C:\Windows\System32\notepad.exe',
+        :compatibility => 1,
+      )[:compatibility]).to eq(:'1')
```
Iristyle: Don't quite understand why value comes back like this.

lib/puppet/type/scheduled_task.rb
```diff
@@ -83,6 +83,15 @@ def insync?(current)
       to determine if a scheduled task is in sync or not."
   end
 
+  newproperty(:compatibility) do
+    desc "The compatibility level associated with the task. May currently only
```
glennsarti: Remove `May`  It's not optional

glennsarti:
There's a bunch of TODOs in the test files still.  Is this PR still WIP?

lib/puppet/type/scheduled_task.rb
Iristyle: Well, specifying the value *is* optional to specify (has to be for backwards compatibility)... but specifying anything other than `1` is not optional at the moment.

Note that this will change / be updated in a later commit as v2 is fleshed out.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -71,12 +71,6 @@ def delete(task_name)
     @tasksched.delete(full_taskname)
   end
 
-  # Execute the current task.
-  #
-  def run
```
michaeltlombardi: I'm assuming these removals indicate that we do _not_ plan to implement them in the future? If we do, leaving them in might not be a big deal / might be useful when we get around to implementing. Otherwise, strong concur with removal of distracting code.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -235,20 +196,14 @@ def trigger(index)
     # The older V1 API uses a starting index of zero, wherease the V2 API uses one.
     # Need to increment by one to maintain the same behavior
     trigger_object = @tasksched.trigger(@definition, index + 1)
-    trigger_object.nil? ? nil : v2trigger_to_v1hash(trigger_object)
+    trigger_object.nil? ? nil : Trigger::V1.from_iTrigger(trigger_object)
```
michaeltlombardi: Appreciate the clarify of the method rename here.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
Iristyle: Yeah, I don't see us implementing these at this point.

michaeltlombardi:
LGTM!

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
glennsarti: History - I added those when this was in Puppet.  As we're now a module, no more reason to carry them through.

PR #14, opened 2018-02-03
(MODULES-6591) Restore v1 type / provider, make v2 provider default for v1 types

Iristyle:
Note that the first commit is a revert of 336fd85, but I reworded the commit message for clarity.

While the implementation of the v1 provider could be gutted at this point to be a mostly empty class, I've decided to hold off on that part for now. That can be addressed at any point in the future (including after the 1.0 release of the module).

This PR should be merged before #13


glennsarti:
The puppet 4.7.1 failure is due to the translations function not existing back.  fastgettext was added in 4.6.0, but might need the stubs from https://github.com/puppetlabs/puppet/commit/129bf9a458b4721995309357e19e589c49f26d2b#diff-9935b4f5c1df75e93b1c7ad923dab468 (Pup 5.2.0) for testing though.

```
undefined method `_' for ...
```

glennsarti:
Decided to up the minimum puppet version to 4.9 instead

appveyor.yml
```diff
@@ -18,7 +18,7 @@ environment:
     RUBY_VER: 24
   - PUPPET_GEM_VERSION: ~> 5.0
     RUBY_VER: 24-x64
-  - PUPPET_GEM_VERSION: 4.7.1
+  - PUPPET_GEM_VERSION: 4.9.0
```
Iristyle: I think we need to change this in `.sync.yml` also, otherwise next modsync will blow it away, right?

https://github.com/puppetlabs/modulesync_configs/blob/master/config_defaults.yml#L123-L138

Will update the commit accordingly...

glennsarti: This is PDKified.  There is no modsync now

glennsarti:
Merged. if we need to modify modsync/pdk we can do that in a new PR

Iristyle:
@glennsarti we will need to revert https://github.com/puppetlabs/modulesync_configs/commit/f56e5875e1f308db10f623a5704c52e56c09cc68 then

What does `.sync.yml` do now? /cc @pmcmaw @DavidS ?

glennsarti:
PDK takes over the sync yaml to do it's pdk convert.  Basically, same purpose, different tool.

glennsarti:
and yes revert required

Iristyle:
modsync removal is at https://github.com/puppetlabs/modulesync_configs/pull/190

PR #16, opened 2018-02-21
(MODULES-6398) Fix error message on invalid day_of_week trigger

glennsarti:
Previously if an invalid value for the day_of_week property was passed (e.g.
MON) it would fail with `Cannot convert "nil" to Fixnum`.  This commit adds
invalid value detection when building the bitmasked value and raises an error
which includes the bad value.  This commit also adds a test for this scenario.


glennsarti:
Error message is now
```
Error: Parameter trigger failed on Scheduled_task[defrag C]: Days_of_week value Sun is invalid. Expected sun, mon, tue, wed, thu, fri or sat. (file: C:/Source/puppetlabs-scheduled_task/scheduled_task/tmp/test.pp, line: 1)
```

PR #17, opened 2018-02-21
(MODULES-6268) Fix error when switching monthly trigger types

glennsarti:
Previously when changing from a monthly day of week to a monthly date trigger
type (and vice versa) Puppet would raise an error.  This was due to Puppet using
the same schedule name (monthly) for two distinct trigger types in windows API.
This commit adds an additional check to the tiggers_same? method which also
evaluates the underlying trigger type, not just the Puppet representation to
determine if two triggers are indeed the same.  This commit also adds tests for
this scenario.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -241,6 +241,7 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
   def triggers_same?(current_trigger, desired_trigger)
     return false unless current_trigger['schedule'] == desired_trigger['schedule']
     return false if current_trigger.has_key?('enabled') && !current_trigger['enabled']
+    return false if translate_hash_to_trigger(desired_trigger)['trigger_type'] != translate_hash_to_trigger(current_trigger)['trigger_type']
 
     desired = desired_trigger.dup
     desired['start_date']  ||= current_trigger['start_date']  if current_trigger.has_key?('start_date')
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -241,6 +241,7 @@ def flush
   def triggers_same?(current_trigger, desired_trigger)
     return false unless current_trigger['schedule'] == desired_trigger['schedule']
     return false if current_trigger.has_key?('enabled') && !current_trigger['enabled']
+    return false if translate_hash_to_trigger(desired_trigger)['trigger_type'] != translate_hash_to_trigger(current_trigger)['trigger_type']
```
Iristyle: I *think* this is an unnecessary conversion for 2 reasons:

* the `'schedule'` value should be setting the `trigger_type`, so those should be directly comparable between `desired_trigger` and `current_trigger` - see https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/master/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb#L297
* We call `translate_hash_to_trigger` a few lines later... so if the first point holds untrue for some reason, probably don't need to build the same object multiple times

Iristyle: After discussing with @glennsarti, realized that in its current state, we do need to perform the conversion to compare because there are 2 different types of "monthly" schedules.

That said, we identified some room for improvement in the overall flow of trigger hash building / initial resource population and will file a new ticket for refactoring some of that, which will introduce some refactoring here as well.

Iristyle: This was finally cleaned up in https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/38 / particularly as part of refactoring `triggers_same?` in https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/38/commits/2e972816b40ef3abceb885d9261152eb8c096922 and https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/38/commits/d7d218ca630fc9f31a620d6fb5842e1da7f8ed3b

PR #18, opened 2018-02-23
(MODULES-6417) Add usage and examples to README

michaeltlombardi:
Prior to this commit the documentation for the `scheduled_task` module
largely pointed back to the official Puppet documentation for information
on the type reference and on the usage of the resource. This commit
ensures that that information now lives in the scheduled_task module and
adds numerous clarifying examples of use.

This will allow future iterative work to continue in this module without
needing to point back to the not-actively-developed type and provider in
core Puppet. Documentation updates to the module can now be safely made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,34 +15,253 @@
 ## Description
 
 This module adds a new [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) provider capable of using the more modern Version 2 Windows API for task management.
+The legacy API does not receive improvements or new features, meaning that if you want to take advantage of improvements to scheduled tasks on Windows you need to use the V2 API.
 
 ## Setup
 
 ### Beginning with scheduled_task
 
-The scheduled_task module adapts the Puppet [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) resource to run using a modern API. To get started, install the module and declare 'taskscheduler_api2' as the provider, for example:
+The scheduled_task module adapts the Puppet [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) resource to run using a modern API.
+To get started, install the module and any existing `scheduled_task` resources will use the V2 API **by default**.
+If you want to continue using the provider for the legacy API you will _need_ to declare that in your manifests.
+For example:
 
 ~~~ puppet
 scheduled_task { 'Run Notepad':
   command  => "notepad.exe",
   ...
-  provider => 'taskscheduler_api2'
+  provider => 'win32_taskscheduler',
 }
 ~~~
 
 ## Usage
 
-See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
+Scheduled tasks are commonly used to kick off a script either once or on a regular cadence.
+In this first example we schedule a cleanup script to run this one time.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup': # Unique name for the scheduled task
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',           # This is the default, but including it is good practice. Flip to 'false' to disable the task.
+  trigger   => [{
+    schedule   => 'once',        # Defines the trigger type; required.
+    start_time => '23:20',       # Defines the time the task should run; required.
+    start_date => '2018-01-01'   # Defaults to the current date; not required.
+  }],
+}
+~~~
+
+If we need to have the cleanup script run every night we can use a daily trigger.
+Just changing the trigger schedule from `once` to `daily` will do the trick.
+Note that we removed the `start_date` from the trigger - it isn't required and, for this task, isn't important.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup Nightly':
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    schedule         => 'daily',
+    start_time       => '23:20'
+  }],
+}
+~~~
+
+You can also set your scheduled tasks to repeat during a set time block.
+Using the cleanup script again, this scheduled task begins at the same time every day and runs once an hour from seven in the morning to seven at night as the SYSTEM account.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup Daily Repeating':
+  ensure    => 'present',
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'         => 'daily',
+    'start_time'       => '07:00',
+    'minutes_duration' => '720',   # Specifies the length of time, in minutes, the task is active
+    'minutes_interval' => '60'     # Causes the task to run every hour
+  }],
+  user      => 'system',           # Specifies the account to run the task as
+}
+~~~
+
+The downside to that task is that it causes the cleanup script to run _every_ day, even on the weekends when there is no activity.
+We can instead use a weekly trigger to fix this:
+
+~~~puppet
+scheduled_task { 'Disk Cleanup Weekly Repeating':
+  ensure    => 'present',
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'         => 'weekly',
+    'start_time'       => '07:00',
+    'day_of_week'      => ['mon', 'tues', 'wed', 'thu', 'fri'], # Note the absence of Sunday and Monday
+    'minutes_interval' => '60',
+    'minutes_duration' => '720'
+  }],
+  user      => 'system',
+}
+~~~
+
+Similarly, we can schedule our cleanup script to run monthly if we decide the cleanup script doesn't need to run as often or is particularly resource intensive.
+The following example sets the scheduled task to run at 0700 on the first day of the month every month:
+
+~~~puppet
+scheduled_task { 'Disk Cleanup Monthly First Day':
+  ensure    => 'present',
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'         => 'monthly',
+    'start_time'       => '07:00',
+    'on'               => [1]        # Run every month on the first day of the month.
+  }],
+  user      => 'system',
+}
+~~~
+
+With the monthly trigger above there is no guarantee that the first day of the month is on a weekend.
+This means that there's a reasonable chance that the script will execute during working hours and impact productivity.
+We can specify the trigger to run on the task on the _first saturday_ of the month instead of the first day:
+
+~~~puppet
+scheduled_task { 'Disk Cleanup Monthly First Saturday':
+  ensure    => 'present',
+  command   => "$::system32\\WindowsPowerShell\\v1.0\\powershell.exe",
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'       => 'monthly',
+    'start_time'     => '07:00',
+    'day_of_week'    => 'sat',     # Specify the day of the week to trigger on
+    'which_occurence => 'first'    # Specify which occurance to trigger on, up to fifth
+  }],
+  user      => 'system',
+}
+~~~
 
 ## Reference
 
 ### Provider
 
+* win32_taskscheduler: This provider manages scheduled tasks on Windows using the legacy API.
 * taskscheduler_api2: Adapts the Puppet scheduled_task resource to use the modern Version 2 API.
 
 ### Type
 
-* scheduled_task: See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
+#### scheduled_task
+
+Installs and manages Windows Scheduled Tasks.
+All attributes except `name`, `command`, and `trigger` are optional; see the description of the [`trigger`](#trigger) attribute for details on setting schedules.
+
+##### `name`
+
+The name assigned to the scheduled task.
+This will uniquely identify the task on the system.
+
+##### `ensure`
+
+The basic property that the resource should be in.
+
+Valid values are `present`, `absent`.
+
+##### `arguments`
+
+Any arguments or flags that should be passed to the command.
+Multiple arguments should be specified as a space-separated string.
+
+##### `command`
+
+The full path to the application to run, without any arguments.
+
+##### `enabled`
+
+Whether the triggers for this task should be enabled.
+This attribute affects every trigger for the task; triggers cannot be enabled or disabled individually.
+
+Valid values are `true`, `false`.
+
+##### `password`
+
+The password for the user specified in the 'user' attribute.
+This is only used if specifying a user other than 'SYSTEM'.
+This parameter will not be used to determine if a scheduled task is in sync or not because there is no way to retrieve the password used to set the account information for a task.
+
+##### `provider`
+
+The specific backend to use for this scheduled_task resource.
+You will seldom need to specify this — Puppet will usually discover the appropriate provider for your platform.
+
+Available providers are:
+
+###### win32_taskscheduler
+
+This provider manages scheduled tasks on Windows using the legacy API.
+
+###### taskscheduler_api2
+
+This provider manages scheduled tasks on Windows using the v2 api.
+
+* Default for `operatingsystem` == `windows`.
+
+##### `trigger`
+
+One or more triggers defining when the task should run.
+A single trigger is represented as a hash, and multiple triggers can be specified with an array of hashes.
+
+A trigger can contain the following keys:
+
+For all triggers:
+
+* `schedule` (Required) — What kind of trigger this is.
+  Valid values are `daily`, `weekly`, `monthly`, or `once`.
+  Each kind of trigger is configured with a different set of keys; see the sections below (once triggers only need a start time/date.)
+* `start_time` (Required) — The time of day when the trigger should first become active.
+  Several time formats will work, but we suggest 24-hour time formatted as HH:MM.
+* `start_date` — The date when the trigger should first become active.
+  Defaults to the current date.
+  You should format dates as YYYY-MM-DD, although other date formats may work (under the hood, this uses Date.parse).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,8 @@ This module adds a new [scheduled_task](https://puppet.com/docs/puppet/latest/ty
 
 ### Beginning with scheduled_task
 
-The scheduled_task module adapts the Puppet [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) resource to run using a modern API. To get started, install the module and declare 'taskscheduler_api2' as the provider, for example:
+The scheduled_task module adapts the Puppet [scheduled_task](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) resource to run using a modern API.
+To get started, install the module and declare 'taskscheduler_api2' as the provider, for example:
```
michaeltlombardi: @Iristyle, @glennsarti:

I wasn't quite sure how to sensibly rewrite this section, but now that the type and provider are included in the module, installing the module clobbers the type/provider in core Puppet, right? And it replaces the default provider with the v2 API provider.

So by default, all you need to do is install the module to get the new provider - if you need to stick with the old provider, you'll need to specify it in your manifests going forward, right?

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+  command => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe -File "C:\\Scripts\Clear-DiskSpace.ps1"',
+  enabled => 'true',
+  trigger => [{
+    schedule         => 'daily',
+    start_time       => '23:20'
+  }],
+}
+~~~
+
+You can also set your scheduled tasks to repeat during an set time block.
```
RandomNoun7: nit, should be 'during a' not an.

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+Re-using the idea of a cleanup script, this scheduled task begins at the same time every day and runs once an hour from seven in the morning to seven at night as the SYSTEM account.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup Daily Repeating':
+  ensure  => 'present',
+  command => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe -File "C:\\Scripts\Clear-DiskSpace.ps1"',
```
RandomNoun7: Is the single slash at end of the file path correct? I don't know if this was a deliberate decision or not, but I won't bother commenting on each instance.

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+In this first example we schedule a cleanup script to run this one time.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup': # Unique name for the scheduled task
+  command => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe -File "C:\\Scripts\Clear-DiskSpace.ps1"',
+  enabled => 'true',             # This is the default, but including it is good practice. Flip to 'false' to disable the task.
+  trigger => [{
+    schedule   => 'once',        # Defines the trigger type; required.
+    start_time => '23:20',       # Defines the time the task should run; required.
+    start_date => '2018-01-01'   # Defaults to the current date; not required.
```
RandomNoun7: Is it proper or not proper to have a trailing comma after the last item in these hashes? I haven't tried it.

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+    'start_time'       => '07:00',
+    'day_of_week'      => ['mon', 'tues', 'wed', 'thurs', 'fri'], # Note the absence of Sunday and Monday
+    'minutes_interval' => '60',
+    'minutes_duration' => '720'
+  }],
+  user    => 'system',
+}
+~~~
+
+Similarly, we can schedule our cleanup script to run monthly if we decide the cleanup script doesn't need to run as often or is particularly resource intensive.
```
RandomNoun7: Do we need to put any effort into speculating on behalf of the user why they might want to run a task monthly? It's such a common requirement that I think we can probably end this scenario description at "to run monthly" or at "need to run as often".

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+##### `ensure`
+
+The basic property that the resource should be in.
+
+Valid values are `present`, `absent`.
+
+##### `arguments`
+
+Any arguments or flags that should be passed to the command.
+Multiple arguments should be specified as a space-separated string.
```
RandomNoun7: Would be useful to add here an example that restates the command above that runs the Clear-DiskSpace script, as a call to powershell.exe with the arguments passed using this parameter?

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+
+##### `command`
+
+The full path to the application to run, without any arguments.
```
RandomNoun7: This contradicts all of the examples above that pass arguments in the command parameter doesn't it?

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+Whether the triggers for this task should be enabled.
+This attribute affects every trigger for the task; triggers cannot be enabled or disabled individually.
+
+Valid values are `true`, `false`.
+
+##### `password`
+
+The password for the user specified in the 'user' attribute.
+This is only used if specifying a user other than 'SYSTEM'.
+Since there is no way to retrieve the password used to set the account information for a task, this parameter will not be used to determine if a scheduled task is in sync or not.
```
RandomNoun7: Consider re-ordering this so that the effect that the user cares about is up front in the sentence, and the justification in the second half. Users skimming docs will often not read entire sentences if the reason they should continue reading it is not immediately apparent.

README.md
```diff
@@ -32,22 +33,231 @@ scheduled_task { 'Run Notepad':
...
+* For monthly (by weekday) triggers:
+  * `months` — Which months the task should run, as an array. Defaults to all months. Each month must be an integer between 1 and 12.
+  * `day_of_week` (Required) — Which day of the week the task should run, as an array with only one element.
+    Each day must be one of `mon`, `tues`, `wed`, `thurs`, `fri`, `sat`, `sun`, or `all`.
+  * `which_occurrence` (Required) — The occurrence of the chosen weekday when the task should run. Must be one of first, second, third, fourth, or fifth.
+
+##### `user`
+
+The user to run the scheduled task as.
+Please note that not all security configurations will allow running a scheduled task as 'SYSTEM', and saving the scheduled task under these conditions will fail with a reported error of 'The operation completed successfully'.
```
RandomNoun7: Is that right? That doesn't sound much like an error message?

README.md
michaeltlombardi: Nah, bug on my part. Good catch!

README.md
michaeltlombardi: This was more part of a narrative doc writing thing, where you take a fictional character through a chain of events with consistent internal logic.

README.md
michaeltlombardi: That's a good idea, I can rework the above examples to use this instead.

README.md
michaeltlombardi: That's because I don't read.

README.md
michaeltlombardi: That's pulled straight from existing docs. 😢

README.md
michaeltlombardi: Fixed.

README.md
michaeltlombardi: Fixed.

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+scheduled_task { 'Disk Cleanup Monthly First Saturday':
+  ensure    => 'present',
+  command   => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'       => 'monthly',
+    'start_time'     => '07:00',
+    'day_of_week'    => 'sat',     # Specify the day of the week to trigger on
+    'which_occurence => 'first'    # Specify which occurance to trigger on, up to fifth
```
ThoughtCrhyme: Is this only available when using the version 2 api? If so, shouldn't we let users know that this example only works on the modern api?

michaeltlombardi: It looks like this is included in the [legacy api](https://puppet.com/docs/puppet/5.4/types/scheduled_task.html#scheduled_task-attribute-trigger) too.

I don't think _at this point_ there's anything net new in the module for the API, it's almost entirely parity / fixes.

README.md
glennsarti: "you'll need to specify it in your manifests going forward, right?"  Correct

glennsarti: Also...bit further up;
```
provider capable of using the more modern Version 2 Windows API for task management
```
Do we need to explain why V2 is better etc?

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
 
 ## Usage
 
-See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
+One common use is to kick off a script, either once or on a regular cadence.
```
glennsarti: nit: Don't say 'One common' and then list multiple options.  Either "Common uses cases are ..." or drop the options

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+In this first example we schedule a cleanup script to run this one time.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup': # Unique name for the scheduled task
```
glennsarti: Not sure how I feel about inline comments.  IMO you show the example and then explain each line later.
If we are leaving inline comments, perhaps align the hash characers?

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+  enabled   => 'true',
+  trigger   => [{
+    schedule         => 'daily',
+    start_time       => '23:20'
+  }],
+}
+~~~
+
+You can also set your scheduled tasks to repeat during a set time block.
+Re-using the idea of a cleanup script, this scheduled task begins at the same time every day and runs once an hour from seven in the morning to seven at night as the SYSTEM account.
```
glennsarti: nit: Reword - "Using the cleanup script again..."

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+~~~puppet
+scheduled_task { 'Disk Cleanup Weekly Repeating':
+  ensure    => 'present',
+  command   => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
+  arguments => '-File "C:\\Scripts\\Clear-DiskSpace.ps1",
+  enabled   => 'true',
+  trigger   => [{
+    'schedule'         => 'weekly',
+    'start_time'       => '07:00',
+    'day_of_week'      => ['mon', 'tues', 'wed', 'thurs', 'fri'], # Note the absence of Sunday and Monday
```
glennsarti: `'thurs'` or '`thu'` ?

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+  trigger   => [{
+    'schedule'         => 'monthly',
+    'start_time'       => '07:00',
+    'on'               => [1]        # Run every month on the first day of the month.
+  }],
+  user      => 'system',
+}
+~~~
+
+One problem with the monthly trigger above is that we can't guarantee that the cleanup script doesn't run during a work day.
```
glennsarti: Double negative here makes it a bit harder to read.  Not sure what to reword to.

README.md
```diff
@@ -32,22 +33,237 @@ scheduled_task { 'Run Notepad':
...
+  * `months` — Which months the task should run, as an array. Defaults to all months. Each month must be an integer between 1 and 12.
+  * `day_of_week` (Required) — Which day of the week the task should run, as an array with only one element.
+    Each day must be one of `mon`, `tues`, `wed`, `thurs`, `fri`, `sat`, `sun`, or `all`.
+  * `which_occurrence` (Required) — The occurrence of the chosen weekday when the task should run. Must be one of first, second, third, fourth, or fifth.
+
+##### `user`
+
+The user to run the scheduled task as.
+Please note that not all security configurations will allow running a scheduled task as 'SYSTEM', and saving the scheduled task under these conditions will fail with a reported error of 'The operation completed successfully'.
+It is recommended that you either choose another user to run the scheduled task, or alter the security policy to allow v1 scheduled tasks to run as the 'SYSTEM' account.
```
glennsarti: `security policy` needs a link to the actual policy to be set.

README.md
michaeltlombardi: Ah, this is a language fail on my part. The _use case_ is to kick off a script, the _implementation details_ are once-only or repeating. At least, that's how I meant it in my head. Will adjust for clarity, thanks!

README.md
michaeltlombardi: I pulled the idea for inline comments from the existing examples on the v1 docs, I'm not married to them. The upside to inline comments is that they're really useful for at-a-glance reference and don't impede the copy-paste usage.

A major downside is they add clutter. I'll at least ensure the hashes align, missed that.

I _think_ most of the stuff covered in the inline comments is covered in the text, either here or in the reference section.

README.md
michaeltlombardi: > With the monthly trigger above there is no guarantee that the first day of the month is on a weekend.
> This means that there's a reasonable chance that the script will execute during working hours and impact productivity.

This is a possible rewrite to clarify the problem. It's longer but explains the actual impact a bit more.

README.md
michaeltlombardi: _Very_ good point, will add.

michaeltlombardi: Actually, my googlefu is failing me here, I can't find a specific security policy that blocks this. Still looking...

michaeltlombardi: It looks like this text was included in the [original type docs](https://github.com/puppetlabs/puppet/commit/0d7c79727cfb158f7182f791ca3ad6e6248c1c23).

README.md
michaeltlombardi: Interestingly, this is apparently supported in the [original implementation](https://github.com/puppetlabs/puppet/commit/0d7c79727cfb158f7182f791ca3ad6e6248c1c23#diff-1df63dec83110755c93b68353185d035R179), even. 😮

README.md
michaeltlombardi: @joshcooper thinks it's related to [this article](https://support.microsoft.com/en-us/help/305612/how-to-prevent-a-user-from-running-task-scheduler-in-windows), possibly preventing Puppet from saving the task in the first place.

michaeltlombardi: Thinking more on this, we can't definitively find a policy that matches up with this text. What we discovered is a policy that, if intentionally set to prevent scheduled tasks from being added/saved, will prevent this resource from working. However, that seems like the policy functioning as intended and isn't the default on any systems.

I've removed this text from the documentation.

README.md
Iristyle: Puppet code allows trailing commas, like Ruby.

README.md
Iristyle: There are some weird behaviors in the v1 API... that hopefully are not the same in the v2 API.

Here is the original comment from @ferventcoder that led to the above - https://github.com/puppetlabs/puppet/pull/2848#issuecomment-49052377
It would make sense to try out his same manifest to see if the behavior is the same.

@glennsarti left a similar comment at https://github.com/puppetlabs/puppet/pull/5042/files#diff-414807a601087eb6bf7e40a0aa442471R229

https://github.com/puppetlabs/puppet/pull/5035/commits also has some details around how `save` worked in v1.

README.md
Iristyle: FWIW, I usually like inline comments... but that's probably a style preference.

README.md
Iristyle: Similar notes for this are in the comment I left at https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/18#discussion_r171111673

README.md
```diff
@@ -15,39 +15,257 @@
...
+  trigger   => [{
+    schedule   => 'once',        # Defines the trigger type; required.
+    start_time => '23:20',       # Defines the time the task should run; required.
+    start_date => '2018-01-01'   # Defaults to the current date; not required.
+  }],
+}
+~~~
+
+If we need to have the cleanup script run every night we can use a daily trigger.
+Just changing the trigger schedule from `once` to `daily` will do the trick.
```
Iristyle: Minor, you changed `trigger` to `daily` *AND* removed `start_date`

README.md
```diff
@@ -15,39 +15,257 @@
...
 
 ## Usage
 
-See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
+Scheduled tasks are commonly used to kick off a script either once or on a regular cadence.
+In this first example we schedule a cleanup script to run this one time.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup': # Unique name for the scheduled task
+  command   => 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
```
Iristyle: Don't want to complicate the example, but it's a little more idiomatic in Puppet to use the [system32 fact](https://puppet.com/docs/facter/3.9/core_facts.html#system32)

Some notes on this at https://puppet.com/docs/puppet/5.3/lang_windows_file_paths.html#file-system-redirection-when-running-32-bit-puppet-on-64-bit-windows

README.md
```diff
@@ -15,39 +15,257 @@
...
+* win32_taskscheduler: This provider manages scheduled tasks on Windows using the legacy API.
 * taskscheduler_api2: Adapts the Puppet scheduled_task resource to use the modern Version 2 API.
 
 ### Type
 
-* scheduled_task: See the [Puppet resource documentation](https://puppet.com/docs/puppet/latest/types/scheduled_task.html) for more information.
+#### scheduled_task
+
+Installs and manages Windows Scheduled Tasks.
+All attributes except `name`, `command`, and `trigger` are optional; see the description of the [`trigger`]() attribute for details on setting schedules.
```
Iristyle: It doesn't appear that this link is working properly....

README.md
```diff
@@ -15,39 +15,257 @@
...
+  Several time formats will work, but we suggest 24-hour time formatted as HH:MM.
+* `start_date` — The date when the trigger should first become active.
+  Defaults to the current date.
+  You should format dates as YYYY-MM-DD, although other date formats may work (under the hood, this uses Date.parse).
+* `minutes_interval` — The repeat interval in minutes.
+* `minutes_duration` — The duration in minutes, needs to be greater than the minutes_interval.
+* For daily triggers:
+  * `every` — How often the task should run, as a number of days.
+    Defaults to 1.
+    “2” means every other day, “3” means every three days, etc.
```
Iristyle: Nit - looks like we've got some weird smartquotes in here.

README.md
michaeltlombardi: Derp, completely missed this!

README.md
```diff
@@ -15,39 +15,258 @@
...
+}
+~~~
+
+You can also set your scheduled tasks to repeat during a set time block.
+Using the cleanup script again, this scheduled task begins at the same time every day and runs once an hour from seven in the morning to seven at night as the SYSTEM account.
+
+~~~ puppet
+scheduled_task { 'Disk Cleanup Daily Repeating':
+  ensure    => 'present',
+  command   => "$system32\\WindowsPowerShell\\v1.0\\powershell.exe",
```
Iristyle: I think we want to use the top scope here like `$::system32`

PR #19, opened 2018-03-15
(WIP) Modules 6526 create v2 provider

michaeltlombardi:


michaeltlombardi:
This is placeholder work on the V2 helper just to make it visible. It will need a lot of cleaning up and rebasing on top of the refactored hash logic.

PR #22, opened 2018-03-22
(MODULES-6844) Add v2-only helper

michaeltlombardi:
These commits add a V2-only helper by cloning and updating the adapter. It adds integration tests for the new helper and strips away pieces of the helper which are for V1 tasks.

It is _out of scope_ to wire this helper up to the provider or otherwise alter the type/provider for this helper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ For all triggers:
   You should format dates as YYYY-MM-DD, although other date formats may work (under the hood, this uses Date.parse).
 * `minutes_interval` — The repeat interval in minutes.
 * `minutes_duration` — The duration in minutes, needs to be greater than the minutes_interval.
+* `random_minutes_interval` — The delay time that is randomly added to the start time of the trigger.
 * For daily triggers:
   * `every` — How often the task should run, as a number of days.
     Defaults to 1.
--- a/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
@@ -1,0 +1,260 @@
+# This class is used to manage V1 compatible tasks using the Task Scheduler V2 API
+# It is designed to be a binary compatible API to puppet/util/windows/taskscheduler.rb but
+# will only surface the features used by the Puppet scheduledtask provider
+#
+require_relative './taskscheduler2'
+require_relative './trigger'
+
+module PuppetX
+module PuppetLabs
+module ScheduledTask
+
+class TaskScheduler2Task
+  # The error class raised if any task scheduler specific calls fail.
+  class Error < Puppet::Util::Windows::Error; end
+
+  public
+  # Returns a new TaskScheduler object. If a work_item (and possibly the
+  # the trigger) are passed as arguments then a new work item is created and
+  # associated with that trigger, although you can still activate other tasks
+  # with the same handle.
+  #
+  # This is really just a bit of convenience. Passing arguments to the
+  # constructor is the same as calling TaskScheduler.new plus
+  # TaskScheduler#new_work_item.
+  #
+  def initialize(work_item = nil, trigger = nil)
+    @tasksched = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2
+
+    new_work_item(work_item, trigger) if work_item && trigger
+  end
+
+  # Returns an array of scheduled task names.
+  #
+  # Emulates V1 tasks by appending the '.job' suffix
+  #
+  def enum
+    @tasksched.enum_task_names(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER,
+      include_child_folders: false,
+      include_compatibility: [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_AT, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V1]).map do |item|
+        @tasksched.task_name_from_task_path(item) + '.job'
+    end
+  end
+  alias :tasks :enum
+
+  # Activate the specified task.
+  #
+  def activate(task_name)
+    raise TypeError unless task_name.is_a?(String)
+    normal_task_name = normalize_task_name(task_name)
+    raise Error.new(_("Scheduled Task %{task_name} does not exist") % { task_name: normal_task_name }) unless exists?(normal_task_name)
+
+    full_taskname = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER + normal_task_name
+
+    @task = @tasksched.task(full_taskname)
+    @full_task_path = full_taskname
+    @definition = @tasksched.task_definition(@task)
+    @task_password = nil
+
+    @task
+  end
+
+  # Delete the specified task name.
+  #
+  def delete(task_name)
+    full_taskname = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER + normalize_task_name(task_name)
+    @tasksched.delete(full_taskname)
+  end
+
+  # Saves the current task. Tasks must be saved before they can be activated.
+  # The .job file itself is typically stored in the C:\WINDOWS\Tasks folder.
+  #
+  # If +file+ (an absolute path) is specified then the job is saved to that
+  # file instead. A '.job' extension is recommended but not enforced.
+  #
+  def save(file = nil)
+    task_object = @task.nil? ? @full_task_path : @task
+    @tasksched.save(task_object, @definition, @task_password)
+  end
+
+  # Sets the +user+ and +password+ for the given task. If the user and
+  # password are set properly then true is returned.
+  #
+  # In some cases the job may be created, but the account information was
+  # bad. In this case the task is created but a warning is generated and
+  # false is returned.
+  #
+  # Note that if intending to use SYSTEM, specify an empty user and nil password
+  #
+  # Calling task.set_account_information('SYSTEM', nil) will generally not
+  # work, except for one special case where flags are also set like:
+  # task.flags = Win32::TaskScheduler::TASK_FLAG_RUN_ONLY_IF_LOGGED_ON
+  #
+  # This must be done prior to the 1st save() call for the task to be
+  # properly registered and visible through the MMC snap-in / schtasks.exe
+  #
+  def set_account_information(user, password)
+    @task_password = password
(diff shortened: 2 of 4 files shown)


michaeltlombardi:
I can't think of much else to rip at this stage from the V2 helper. Some of the logic lives in the trigger class and the indexing stuff is in the provider, both of which are out of scope for this ticket.

glennsarti:
Should there be unit tests for this or integration only?

michaeltlombardi:
@glennsarti Possibly - I didn't see any unit tests for the adapter code, so I didn't add any new ones. I can noodle on adding some, but my original thought was to update the provider unit tests in the ticket to wire up the new helper to the provider.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -259,6 +259,7 @@ def self.append_v1trigger(definition, v1trigger)
       end
 
       # Values for all Trigger Types
+      iTrigger.RandomDelay         = "PT#{v1trigger['random_minutes_interval']}M" unless v1trigger['random_minutes_interval'].nil?   || v1trigger['random_minutes_interval'].zero?
```
Iristyle: Could you fill in the docs on this commit as well.  There's nothing anywhere that mentions a user may set this.

michaeltlombardi: Prior to this being hooked up with the provider, they definitely cannot. Where would that doc update land in this workset? Documenting it in this helper seems like the wrong place but we've pushed off updating the type/provider in this ticket for the wire-up ticket.

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    result = `#{full_cmd}`
+
+    result.strip
+  end
+
+  match do |actual|
+    from_ps = run_ps(ps_cmd)
+
+    # This matcher probably won't tolerate UTF8 characters
```
Iristyle: You can remove that comment... `==` should work fine for Unicode(*).


(*) There are some cases where Ruby can have problems - but its typically only when comparing normalized to non-normalized Unicode, which you wouldn't have to worry about here. See http://blog.honeybadger.io/ruby-s-unicode-support/

Iristyle: Also, if you're concerned about the output coming back from PS itself ... multiple responses would come back as the same sequence of bytes (if corrupted by capturing things over `stdout`), so should also match for that reason.

Iristyle:
I think we might want to add the commit that exposes the task compatibility and then use that in tests?  Something similar to https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/19/commits/9e10ee5fd06d438e2510c1a441bba1eb130fe6b8?

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
michaeltlombardi: Okay, I pulled this comment over when I stole the method itself from Glenn's work on the adapter integration tests.

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      'type'         => {
+        'weeks_interval' => 1,
+        'days_of_week'   => Win32::TaskScheduler::MONDAY,
+      },
+      'trigger_type'            => :TASK_TIME_TRIGGER_WEEKLY,
+    }),
+    defaults.merge({
+      'type'         => {
+        'months' => Win32::TaskScheduler::JANUARY,
+        # TODO: this value seems to be cooked up wrong
```
Iristyle: We still have this test TODO here which should probably be removed / sorted out...

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      'trigger_type'            => :TASK_TIME_TRIGGER_ONCE,
+    }),
+    defaults.merge({
+      'type'                    => { 'days_interval' => 1 },
+      'trigger_type'            => :TASK_TIME_TRIGGER_DAILY,
+    }),
+    defaults.merge({
+      'type'         => {
+        'weeks_interval' => 1,
+        'days_of_week'   => Win32::TaskScheduler::MONDAY,
```
Iristyle: I believe this constant now lives in the Trigger class... but might not yet until my next Trigger PR goes in

Iristyle:
Do we wan to modify https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/master/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb#L124 so that it also uses `TaskScheduler2Task` in addition to `TaskScheduler2V1Task`... I think that as we discussed, this can be deferred until the new helper is integrated with the provider.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
```diff
@@ -36,7 +36,7 @@ def initialize(work_item = nil, trigger = nil)
   def enum
     @tasksched.enum_task_names(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER,
       include_child_folders: false,
-      include_compatibility: [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_AT, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V1]).map do |item|
+      include_compatibility: [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V2]).map do |item|
```
Iristyle: This is probably the commit where we want some tests to at least vet the compatibility on the task returned.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
Iristyle: Yeah, that's a good point.  Of course, users won't really get their hands on the module until the code actually ships, at which point the code will be integrated.

I don't feel super strongly either way, but it's usually helpful to update docs when a code change lands that enables a feature?

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
michaeltlombardi: They do not, but we can be sure to update once the constants move.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
michaeltlombardi: Addressed with a minor readme update:

README.md
```diff
@@ -228,6 +228,7 @@ For all triggers:
   You should format dates as YYYY-MM-DD, although other date formats may work (under the hood, this uses Date.parse).
 * `minutes_interval` — The repeat interval in minutes.
 * `minutes_duration` — The duration in minutes, needs to be greater than the minutes_interval.
+* `random_minutes_interval` — The delay time that is randomly added to the start time of the trigger.
```
glennsarti: Is the default 0?

glennsarti: Can we use 0?

michaeltlombardi: Yes, the default is zero. I haven't tried to set it to zero (the default value is actually an empty string which it treats as a zero). Will test that now!

michaeltlombardi: Ensured the unit tests verify setting zero - it turns out that if you try setting it to zero it returns an empty string.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
```diff
@@ -0,0 +1,260 @@
+# This class is used to manage V1 compatible tasks using the Task Scheduler V2 API
```
Iristyle: This blurb is not right upon the 2nd commit and should get a minor update to reflect that.

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    it 'should be able to update a command' do
+      new_application_name = 'notepad.exe'
+      ps_cmd = '[string]((Get-ScheduledTask | ? { $_.TaskName -eq \'' + @task_name + '\' }).Actions[0].Execute)'
+      task = subject.activate(@task_name)
+
+      expect('cmd.exe').to be_same_as_powershell_command(ps_cmd)
+
+      subject.application_name = new_application_name
+      subject.save
+
```
Iristyle: whitespace

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    it 'should be able to update command parameters' do
+      new_parameters = '/nonsense /utter /nonsense'
+      ps_cmd = '[string]((Get-ScheduledTask | ? { $_.TaskName -eq \'' + @task_name + '\' }).Actions[0].Arguments)'
+      task = subject.activate(@task_name)
+
+      expect('/c exit 0').to be_same_as_powershell_command(ps_cmd)
+
+      subject.parameters = new_parameters
+      subject.save
+
```
Iristyle: remove whitespace

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    it 'should be able to update the working directory' do
+      new_working_directory = 'C:\Somewhere'
+      ps_cmd = '[string]((Get-ScheduledTask | ? { $_.TaskName -eq \'' + @task_name + '\' }).Actions[0].WorkingDirectory)'
+      task = subject.activate(@task_name)
+
+      expect('').to be_same_as_powershell_command(ps_cmd)
+
+      subject.working_directory = new_working_directory
+      subject.save
+
```
Iristyle: remove whitespace

PR #24, opened 2018-03-29
(MODULES-6526) Add compatibility in task type

michaeltlombardi:
- Report on the compatibility level of tasks enumerated from the
   system.

   This requires adding a value to the scheduled_task type named
   compatibility. Based on how the provider works, it currently only
   enumerates v1 tasks, and therefore only returns 1.

   Since the property may now also be set in a manifest, limit the
   value to 1 so that it cannot be set improperly. Default the value
   to 1 to preserve backward compatibility.

   `puppet resource` will now emit the value as well

   A future commit will add compatibility versioning for V2 when
   that helper is wired up.

### DIFF
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -83,6 +83,15 @@ Puppet::Type.newtype(:scheduled_task) do
       to determine if a scheduled task is in sync or not."
   end
 
+  newproperty(:compatibility) do
+    desc "The compatibility level associated with the task. May currently only
+      be set to 1 for compatibility with tasks on a Windows XP or Windows Server
+      2003 computer."
+
+    newvalue(1)
+    defaultto(1)
+  end
+
   newproperty(:trigger, :array_matching => :all) do
     desc <<-'EOT'
       One or more triggers defining when the task should run. A single trigger is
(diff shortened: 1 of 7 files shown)


lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
```diff
@@ -176,6 +176,10 @@ def trigger_count
     @tasksched.trigger_count(@definition)
   end
 
+  def compatibility
+    @tasksched.compatibility(@definition)
```
Iristyle: As warty as this is, at least its consistent, and we can clean it up later... 👍

Iristyle:
Did you run a sample manifest through with compatibility set to `1`? How about other values?

lib/puppet/type/scheduled_task.rb
```diff
@@ -83,6 +83,15 @@ def insync?(current)
       to determine if a scheduled task is in sync or not."
   end
 
+  newproperty(:compatibility) do
```
Iristyle: I think we should update the README in this commit to document this.

Do we want to add the relevant bit to `CHANGELOG` too?

michaeltlombardi: Adding both is a good idea.

PR #25, opened 2018-04-04
(MODULES-6526) Add setter for compatibility

michaeltlombardi:
Prior to this commit the code was added to enable
users to specify a compatibility version to create
a task. Unfortunately, there was no testing for
verifying idempotency. When run against an existing
task the resource would fail because it can not
handle the compatibility property.  This commit
ensures that the compatibility property is settable.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -164,6 +164,10 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
     end
   end
 
+  def compatibility=(value)
+    task.compatibility = value
+  end
+
   def trigger=(value)
     desired_triggers = value.is_a?(Array) ? value : [value]
     current_triggers = trigger.is_a?(Array) ? trigger : [trigger]
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -90,6 +90,14 @@ Puppet::Type.newtype(:scheduled_task) do
 
     newvalue(1)
     defaultto(1)
+
+    validate do |value|
+      raise Puppet::Error.new(_("must be a number")) unless value.is_a?(Integer)
+      super(value)
+    end
+
+    # override default munging of newvalue() to symbol, treating input as number
+    munge { |value| value }
   end
 
   newproperty(:trigger, :array_matching => :all) do
--- a/spec/unit/puppet/type/scheduled_task_spec.rb
+++ b/spec/unit/puppet/type/scheduled_task_spec.rb
@@ -114,7 +114,20 @@ describe Puppet::Type.type(:scheduled_task), :if => Puppet.features.microsoft_wi
         :title        => 'Foo',
         :command      => 'C:\Windows\System32\notepad.exe',
         :compatibility => 1,
-      )[:compatibility]).to eq(:'1')
+      )[:compatibility]).to eq(1)
+    end
+
+    it 'should not allow the string value "1"' do
+      expect {
+        described_class.new(
+          :name         => 'Foo',
+          :command      => 'C:\Windows\System32\notepad.exe',
+          :compatibility => "1"
+        )
+      }.to raise_error(
+        Puppet::ResourceError,
+        /Parameter compatibility failed on Scheduled_task\[Foo\]: must be a number/
+      )
     end
 
     it 'should not allow 2' do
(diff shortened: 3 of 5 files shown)


lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -164,6 +164,11 @@ def enabled=(value)
     end
   end
 
+  def compatibility=(value)
+    # It tries to send :"1" as the value. This works but is clearly wrong.
+    task.compatibility = value.to_s.to_i
```
Iristyle: Can we trace this back a bit further before proceeding with this?

michaeltlombardi: Definitely, I meant to ask earlier if you had any idea why it might give that value. Is it possibly because of the newvalue in the type declaration?

lib/puppet/type/scheduled_task.rb
```diff
@@ -90,6 +90,10 @@ def insync?(current)
 
     newvalue(1)
     defaultto(1)
+
+    munge do |setting|
```
michaeltlombardi: @Iristyle I haven't figured out _why_ yet but simply adding a munge here - not acting on the value at all, just returning it - is enough to prevent the weird conversion to symbol/string.

I'm going to see if I can discover anything about this behavior.

michaeltlombardi:
This passes adhoc now, but will still catch fire on non-UTC systems.

spec/unit/puppet/type/scheduled_task_spec.rb
```diff
@@ -114,7 +114,7 @@
         :title        => 'Foo',
         :command      => 'C:\Windows\System32\notepad.exe',
         :compatibility => 1,
-      )[:compatibility]).to eq(:'1')
```
michaeltlombardi: With the munging in place, this had to be fixed. Do you remember where this value came from / why it happens, @Iristyle?

Is it related to [this note from the docs](
https://puppet.com/docs/puppet/5.3/custom_types.html#properties)?

> If you specify valid values, then Puppet will only accept those values, and it will automatically handle accepting either strings or symbols.

Iristyle: https://puppet.com/docs/puppet/5.5/custom_types.html#validation-and-munging

> The default validate method looks for values defined using newvalues and if there are any values defined it accepts only those values (this is exactly how allowed values are validated). The default munge method converts any values that are specifically allowed into symbols. If you override either of these methods, note that you lose this value handling and symbol conversion, which you’ll have to call super for.
> 
> Values are always validated before they’re munged.
> 
> Lastly, validation and munging only* happen when a value is assigned. They have no role to play at all during use of a given value, only during assignment.

Iristyle: A lot of the `newvalue` tests are in https://github.com/puppetlabs/puppet/blob/master/spec/unit/property_spec.rb#L230-L397

So yes, the default munge will `to_symbol` the values instead of allowing them to remain as integers, which is something I did not previously realize.  It's probably more idiomatic Puppet then to stick with an integer definition and a `validate` block that tests the range, rather than the enumerated values... but I'm going to test a few things locally.

lib/puppet/type/scheduled_task.rb
Iristyle: Ok, so this ends up being really close, but we probably want to tweak it slightly. My preference would be for the former unless you have strong feelings about the alternative.

2 options:

#### Puppet 4+ solution

More strictly adhere to the value being a Puppet [number](https://puppet.com/docs/puppet/5.3/lang_data_number.html), and make the contract with users only allow `compatibility => 1` in a manifest

```ruby
validate do |value|
  raise Puppet::Error.new(_("compatibility must be a number")) unless value.is_a?(Integer)
end

# override default munging of newvalue() to symbols, treating user input as number
munge { |value| value }
```

Given validation *only* allows the given `Integer` type, and `newvalues` only allows `1`, this makes for a pretty clear enforcement.

#### Puppet 3+ solution

Accept either `compatibility => 1` OR `compatibility => '1'`

```ruby
# override default munging of newvalue() to symbols, allowing string or numeric values
munge do |value|
  if value.is_a?(String) && value =~ /^[-0-9]+$/
    Integer(value)
  else
    value
  end
end
```

FWIW, existing types and providers tend to take the latter approach, but there are a couple of reasons for this:

* It ensures that old Puppet 3 code generally still works on Puppet 4
* Nobody cared enough to make things more strict (especially in light of the first point)

I don't think either of the above points are important for net new code at this point in time. This munge also allows all types other than `String` through the `else` (like `Array`), which also makes the first variation better.

Iristyle:
@michaeltlombardi I've opened up a PR against your PR at https://github.com/michaeltlombardi/puppetlabs-scheduled_task/pull/1

michaeltlombardi:
@Iristyle updated.

Iristyle:
NOTE: This builds on #24 which was incomplete. We later, due to insufficient testing, found another issue which was resolved in #30

PR #26, opened 2018-04-20
(WIP) V2 Helper Wiring

michaeltlombardi:
This code is currently in an exploratory state to wire up the new V2 helper.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -2,6 +2,7 @@ require 'puppet/parameter'
 
 if Puppet.features.microsoft_windows?
   require File.join(File.dirname(__FILE__), '../../../puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task')
+  require File.join(File.dirname(__FILE__), '../../../puppet_x/puppetlabs/scheduled_task/taskscheduler2_task')
 end
 
 Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
@@ -12,8 +13,16 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
   defaultfor :operatingsystem => :windows
   confine    :operatingsystem => :windows
 
+  def task_scheduler_helper
+    resource[:compatibility] == 1 ? PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task : PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task
+  end
+
   def self.instances
-    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new.tasks.collect do |job_file|
+    # Works fine
+    # PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new.tasks.collect do |job_file|
+    # Explodes for some reason, see win32_taskscheduler_spec.rb:696
+    # Something in that test doesn't like the enum for the v2 helper but is fine with the v1 adapter
+    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new.tasks.collect do |job_file|
       job_title = File.basename(job_file, '.job')
       new(
         :provider => :taskscheduler_api2,
@@ -23,13 +32,13 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
   end
 
   def exists?
-    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new.exists? resource[:name]
+    task_scheduler_helper.new.exists? resource[:name]
   end
 
   def task
     return @task if @task
 
-    @task ||= PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new
+    @task ||= task_scheduler_helper.new
     @task.activate(resource[:name] + '.job') if exists?
 
     @task
@@ -219,7 +228,7 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
 
   def create
     clear_task
-    @task = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new(resource[:name], dummy_time_trigger)
+    @task = task_scheduler_helper.new(resource[:name], dummy_time_trigger)
     self.command = resource[:command]
 
     [:arguments, :working_dir, :enabled, :trigger, :user].each do |prop|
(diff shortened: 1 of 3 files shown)


michaeltlombardi:
Needs fixing, testing, and rebasing.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -2,6 +2,7 @@
 
 if Puppet.features.microsoft_windows?
   require File.join(File.dirname(__FILE__), '../../../puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task')
+  require File.join(File.dirname(__FILE__), '../../../puppet_x/puppetlabs/scheduled_task/taskscheduler2_task')
```
michaeltlombardi: Making sure both the helper and the adapter are loaded

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -12,8 +13,16 @@
   defaultfor :operatingsystem => :windows
   confine    :operatingsystem => :windows
 
+  def task_scheduler_helper
+    resource[:compatibility] == 1 ? PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task : PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task
```
michaeltlombardi: Not sure this is an elegant or sane solution, because this check will get called every time `task_scheduler_helper` is used, but it's a quick comparison.

lib/puppet/type/scheduled_task.rb
```diff
@@ -89,6 +89,7 @@ def insync?(current)
       2003 computer."
 
     newvalue(1)
+    newvalue(2)
```
michaeltlombardi: Ensure the new compatibility mode is allowed. This will require `/spec/unit/puppet/type/scheduled_task_spec.rb:134` to be updated now that the value is allowed.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
```diff
@@ -31,7 +31,7 @@ def initialize(work_item = nil, trigger = nil)
   def enum
     @tasksched.enum_task_names(PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::ROOT_FOLDER,
       include_child_folders: false,
-      include_compatibility: [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V2]).map do |item|
+      include_compatibility: [PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V2, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_COMPATIBILITY_V1]).map do |item|
```
michaeltlombardi: Was investigating having this return all available tasks instead of just V2, and therefore allowing `self.instances` in the provider to correctly populate with all manageable scheduled tasks on the system.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -12,8 +13,16 @@
...
+  end
+
   def self.instances
-    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new.tasks.collect do |job_file|
+    # Works fine
+    # PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new.tasks.collect do |job_file|
+    # Explodes for some reason, see win32_taskscheduler_spec.rb:696
+    # Something in that test doesn't like the enum for the v2 helper but is fine with the v1 adapter
+    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new.tasks.collect do |job_file|
```
michaeltlombardi: Getting this to work properly is the last major hurdle to implementation of the wiring (tests probably still needed elsewhere) - as is in master, this will only return legacy jobs. For some reason, pointing to the jobs collection from the new helper blows up. Still investigating but expect it to have something to do with the legacy adapter appending '.job' to the end of job names and the V2 helper not bothering to do so.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -23,13 +32,13 @@ def self.instances
...
-    PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new.exists? resource[:name]
+    task_scheduler_helper.new.exists? resource[:name]
   end
 
   def task
     return @task if @task
 
-    @task ||= PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new
+    @task ||= task_scheduler_helper.new
     @task.activate(resource[:name] + '.job') if exists?
```
michaeltlombardi: Possible ongoing issues with the appending of '.job'

Iristyle:
This is superseded by #27 ... which addresses all of the above problems.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
Iristyle: I think in this case it was just because tests weren't updated from `TaskScheduler2V1Task` to `TaskScheduler2Task`... see https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/27/commits/b73f8c3282401f9575ba4208875c283f1c34583c

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
Iristyle: Yeah, think it was just the tests.

Still need to do some more verification / validation of the PR at https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/27, but at least specs are passing

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
Iristyle: Yeah, we should discuss further, but in the new PR I just replaced `TaskScheduler2V1Task` with `TaskScheduler2Task` everywhere, and I added `AT` in addition to `V1` like you've done here.

That completely removes the adapter from the workflow, which now makes me wonder why we needed `TaskScheduler2Task` class in the first place. I think removing all the Trigger related code has made the picture a lot clearer now.

lib/puppet/type/scheduled_task.rb
Iristyle: Done in https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/27/commits/8373a8897bc9e22c2bc43c0086a67bc8eae76efd

PR #29, opened 2018-04-26
(MODULES-7071) Remove random_minutes_interval

Iristyle:
- random_minutes_interval is a completely unsupported property in the
   V1 API. It is unusable / always set to 0 in the struct used by
   the COM APIs.

 - Remove all references to this name (including from README) so that
   users are not directed to use this setting.

   NOTE: One instance had to be left in the Trigger helpers for the
   sake of continuity / passing tests. The tests call directly into
   Win32::TaskScheduler, which is implemented inside taskscheduler.rb
   which lives in the Puppet core code. That code produces a hash with
   random_minutes_interval as it is translating from the COM struct.
   Since this cannot be changed, and since the specs found in
   taskscheduler2_v1task_spec compare triggers produced by the
   adapter class to the original, the helper method V1.from_iTrigger
   continues to produce random_minutes_interval, hardcoded to 0, just
   like the original.

 - The V2 API has a RandomDelay property, but is not the same as the
   random_minutes_interval. RandomDelay is specified as an ISO8601
   duration style string like PnYnMnDTnHnMnS. This will later be
   surfaced as a new trigger attribute in the future.


lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -62,6 +62,12 @@ def user
     account
   end
 
+  # TODO: this has to be supplied to still use the old provider for managing tasks
+  # technically needed as a bug fix
+  def compatibility
```
Iristyle: Need to compare to original commit that added compatibility to make sure everything is added as necessary here.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -357,7 +356,9 @@ def self.from_iTrigger(iTrigger)
         'minutes_duration'        => Duration.to_minutes(iTrigger.Repetition.Duration),
         'minutes_interval'        => Duration.to_minutes(iTrigger.Repetition.Interval),
         'flags'                   => trigger_flags,
-        'random_minutes_interval' => Trigger.string_to_int(iTrigger.Randomdelay)
+        # the V1 COM API always produces a value of 0 here and this is kept for
```
Iristyle: This is the one reference I left @michaeltlombardi

michaeltlombardi: 👍

Iristyle:
This has at least passed adhoc at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/All/job/forge-windows_puppetlabs-scheduled_task_i18n-ruby_adhoc/8/ but should undergo additional manual validation locally.

PR #30, opened 2018-04-26
(MODULES-6526) Confine compatibility to a feature

michaeltlombardi:
Prior to this commit any calls to the win32 legacy provider after
the introduction of the compatibility setting cause a run to blow
up because that provider does not have a way to manage compatibility.

This commit adds a feature, `:compatibility`, and limits the methods
for compatibility setting and getting to that feature.  Note that this
method will cause the legacy provider and any provider that does not
support the feature to silently ignore compatibility if passed in a
manifest.  It will not error but it also won't do anything.

If swapping between providers, this can probably cause some strange
behavior.  All net-new features and properties for this module will
probably need to be added this way to prevent explosions in the
legacy provider.

The added acceptance test is a simple copy-change and is not robust.
A future ticket and PR should ensure that the acceptance suite does
a bit more to test against both providers.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -10,6 +10,8 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
   defaultfor :operatingsystem => :windows
   confine    :operatingsystem => :windows
 
+  has_feature :compatibility
+
   def self.instances
     PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task.new.tasks.collect do |job_file|
       job_title = File.basename(job_file, '.job')
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -7,6 +7,10 @@ Puppet::Type.newtype(:scheduled_task) do
     except `name`, `command`, and `trigger` are optional; see the description
     of the `trigger` attribute for details on setting schedules."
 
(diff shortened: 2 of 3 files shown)


michaeltlombardi:
The new acceptance test was just a hack to push something through adhoc for now - seems to be passing:

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/scheduled_task/job/forge-windows_puppetlabs-scheduled_task_intn-sys_smoke-adhoc/lastSuccessfulBuild/

michaeltlombardi:
Our acceptance tests in general will need some reworking for testing both providers, but I think we can largely punt on that to a future ticket to properly address the problem.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -10,6 +10,8 @@
   defaultfor :operatingsystem => :windows
   confine    :operatingsystem => :windows
 
+  has_feature :compatibility
```
Iristyle: I think this feature name is global, and should probably be something like `scheduled_task_compatibility` or similar to prevent collision.

Iristyle: Strike that, it appears that I'm mistaken and these feature names are specific to the type.

michaeltlombardi:
Latest revision has passed adhoc:  https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/scheduled_task/job/forge-windows_puppetlabs-scheduled_task_intn-sys_full-pe-current-adhoc/lastSuccessfulBuild/

PR #31, opened 2018-04-27
(MODULES-6895) Additional Trigger conversion bugfixes and refactorings

Iristyle:
This should be merged prior to #27, partly because some code is being extracted from `taskscheduler_api2` and that code is about to be copied into `win32_taskscheduler`.  Therefore it makes more sense to do this work first, especially given it fixes some critical bugs in converting `ITrigger` back to a V1 internal representation.

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -405,7 +405,10 @@ module Trigger
           'weeks_interval' => 1
         }
       when 'monthly'
-        { 'months' => Month.indexes_to_bitmask((1..12).to_a) }
+        {
+          'months' => Month.indexes_to_bitmask((1..12).to_a),
+          'days' => 0
+        }
       end
     end
 
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -564,6 +564,43 @@ module Trigger
       trigger
     end
 
+    def self.to_manifest_hash(v1trigger)
+      unless V2::V1_TYPE_MAP.keys.include?(v1trigger['trigger_type'])
+        raise ArgumentError.new(_("Unknown trigger type %{type}") % { type: v1trigger['trigger_type'] })
+      end
+
+      manifest_hash = {}
+
+      case v1trigger['trigger_type']
+      when :TASK_TIME_TRIGGER_DAILY
+        manifest_hash['schedule'] = 'daily'
+        manifest_hash['every']    = v1trigger['type']['days_interval'].to_s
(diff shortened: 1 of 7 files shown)


lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -334,25 +334,41 @@ def self.canonicalize_and_validate(hash)
       hash
     end
 
+    V2_SUPPORTED_INTERFACE_MAP =
```
michaeltlombardi: Strong prefer on using this instead of mapping in the case.

Iristyle: Yeah, even better... we already has this defined in `V2` ... updated to use that instead.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -154,6 +154,114 @@
...
+        :RunOnLastWeekOfMonth => true,
+        :RandomDelay  => 'P2DT5S',
+      },
+    ].each do |trigger_details|
+      it "should convert a full #{trigger_details[:ole_type]} expectedly" do
+        iTrigger = FILLED_ITRIGGER_PROPERTIES.merge(trigger_details)
+        # stub is not usable outside of specs (like in DEFAULT_ITRIGGER_PROPERTIES)
+        iTrigger[:Repetition] = stub(iTrigger[:Repetition])
+        iTrigger = stub(iTrigger)
+        expect(subject.class.from_iTrigger(iTrigger)).to_not be_nil
```
Iristyle: `to_not be_nil` is not very .. uhh.. useful. Updating this momentarily

Iristyle:
One more commit to add to this one that does the refactor I'm most interested in... had to clean some cruft out of the way first

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -154,6 +154,164 @@
...
+        {
+          'trigger_type' => :TASK_TIME_TRIGGER_MONTHLYDOW,
+          'type' => { 'weeks' => 0b11111, 'days_of_week' => 0b1111111, 'months' => 0b111111111111 },
+        }
+      },
+    ].each do |trigger_details|
+      it "should convert a full #{trigger_details[:iTrigger][:ole_type]} expectedly" do
+        iTrigger = FILLED_ITRIGGER_PROPERTIES.merge(trigger_details[:iTrigger])
+        # stub is not usable outside of specs (like in DEFAULT_ITRIGGER_PROPERTIES)
+        iTrigger[:Repetition] = stub(iTrigger[:Repetition])
```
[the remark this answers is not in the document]
glennsarti: Logged MODULES-7082 to update this.

PR #32, opened 2018-04-27
(MODULES-7026) Ensure timezones other than UTC work

michaeltlombardi:
Prior to this commit the code for the trigger logic under taskcheduler_api2 only
ever worked for machines whose timezone was set to UTC - it always returned the
datetime string from `iso8691_datetime` with no UTC offset.  This commit updates
that method to return a localized datetime string including the appropriate
offset.

It also updates the unit tests since local offset can vary from system to system.
It does not add any integration or acceptance tests.

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -71,8 +71,8 @@ module Trigger
   end
   module_function :string_to_date
 
-  def iso8601_datetime(year, month, day, hour, minute)
-    DateTime.new(year, month, day, hour, minute, 0).iso8601
+  def iso8601_datetime(year, month, day, hour, minute, offset = DateTime.now.offset)
+    DateTime.new(year, month, day, hour, minute, 0, offset).iso8601
   end
   module_function :iso8601_datetime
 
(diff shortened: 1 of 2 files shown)


lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -72,7 +72,8 @@ def string_to_date(value)
   module_function :string_to_date
 
   def iso8601_datetime(year, month, day, hour, minute)
```
Iristyle: I wonder if it would be better to pass the timezone with a default in here as the last parameter? I think you can call  a `new_offset` method per https://ruby-doc.org/stdlib-2.3.1/libdoc/date/rdoc/DateTime.html and just automatically use the current timezone info.

I suspect then you don't need to change the tests (untested locally - just something to think about)

michaeltlombardi: Will a user ever choose a TZ other than local/utc? I'm trying to think of a situation where I would specify another timezone.

Iristyle: I don't think so - but it may be useful when writing tests. It could be inlined into the method as well.

michaeltlombardi:
Close unneeded.

PR #33, opened 2018-04-30
(MODULES-7026) Ensure timezones other than UTC work

Iristyle:
Prior to this commit the code for the trigger logic under taskcheduler_api2 only
ever worked for machines whose timezone was set to UTC - it always returned the
datetime string from `iso8691_datetime` with no UTC offset.  This commit updates
that method to return a localized datetime string including the appropriate
offset.

It also updates the unit tests since local offset can vary from system to system.
It does not add any integration or acceptance tests.


Iristyle:
We might want to use `time.local` and `time.utc` in a few cases per https://stackoverflow.com/questions/5941613/are-the-date-time-and-datetime-classes-necessary

glennsarti:
This seems to make sense.  Going to be interesting to see the result as the DST barrier is crossed.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -62,17 +64,20 @@ def string_to_int(value)
   end
   module_function :string_to_int
 
-  def string_to_date(value)
+  # TODO: rename this to imply better how its parsed?
+  def string_to_date(value, convert_to = :utc)
```
Iristyle: @michaeltlombardi any thoughts about this always returning UTC? i.e. rename it to `iso8601_datetime_to_utc` and remove the `convert_to` param? Don't think we need the param for tests... and its really the only code path.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -62,17 +64,20 @@ def string_to_int(value)
...
-    DateTime.parse(value)
+    # defaults to parsing as local with no timezone passed
+    time = Time.parse(value)
+    convert_to == :utc ? time.getutc : time
   end
   module_function :string_to_date
 
-  def iso8601_datetime(year, month, day, hour, minute)
-    DateTime.new(year, month, day, hour, minute, 0).iso8601
+  def iso8601_datetime(year, month, day, hour, minute, format = :local)
```
Iristyle: Add docs for `format` allowing `:local` and `:utc` as values.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -38,7 +38,17 @@
       { :input => '2018-01-02T03:04:05', :expected => DateTime.new(2018, 1, 2, 3, 4, 5) },
       { :input => '1899-12-30T00:00:00', :expected => DateTime.new(1899, 12, 30, 0, 0, 0) },
     ].each do |value|
-      it "should return a valid DateTime object for date string #{value[:input]}" do
+      it "should return a valid DateTime object for date string #{value[:input]} with an empty offset" do
```
Iristyle: No longer a `DateTime`

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
michaeltlombardi: Yeah, this makes sense to me.  It's a bit more descriptive, too.

spec/integration/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task_spec.rb
```diff
@@ -39,6 +39,9 @@
       expect(subjectv2.parameters).to eq(subjectv1.parameters)
       expect(subjectv2.application_name).to eq(subjectv1.application_name)
       expect(subjectv2.trigger_count).to eq(subjectv1.trigger_count)
+      # NOTE: this currently only works on a system with UTC dates because
```
Iristyle: This may mean that V1 hash should use local time instead of UTC. Will have to revisit prior to merge.

Iristyle:
@michaeltlombardi could you spot check the first and second commits ... if it looks good / makes sense, I will squash them and reword commit message accordingly to match.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -553,16 +549,16 @@ def self.from_manifest_hash(manifest_hash)
...
       end
 
-      if start_date = manifest_hash['start_date']
-        start_date = Date.parse(start_date)
-        trigger['start_year']  = start_date.year
-        trigger['start_month'] = start_date.month
-        trigger['start_day']   = start_date.day
-      end
+      # manifests specify datetime in the local timezone, same as V1 trigger
+      # TODO: figure out how 0 0 0 manifests... OR... just check and pass along?
```
Iristyle: What's being done here passes tests.. but any thoughts on the above question?

michaeltlombardi: Personally prefer to check and pass, not sure anyone _should_ be passing all zeroes.

Iristyle: The V1 API allows 0s... if it wasn't clear, this is the method that takes an ITrigger and converts it to the equivalent V1 hash. When a `StartBoundary` comes in with the value `''` the question is ... how would that be stored?

It probably doesn't matter much in the long term as this conversion will go the way of the dodo once the other PR is merged.

michaeltlombardi: Oh, I thought this was converting from the hash to the ITrigger.  The comment lines _seem_ to live inside of https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/33/files#diff-ede2463e358119c5b73d7d80617b78d2L514

Iristyle: I actually just realized (after debugging some more to resolve this point) that I completely screwed something up - values for `start_year`, `start_month` and `start_day` are set in the hash for the tests, but we're looking at `start_date` only now because I misinterpreted the earlier code.

The original code allowed *both* a `start_date` value and individual values for the constituent pieces, allowing `start_date` (when specified) to take precedence (which is how things are written in a manifest).  I don't really understand why we allow this ambiguity, but we should probably file a ticket to take a quick audit (probably someone in the past got confused about manifest hash vs V1 esque hash).

Fixing to follow old behavior for now, but I think we have a few additional things to do later.

michaeltlombardi: Strong preference for date over the weird year/month/day individual specification, especially for humans.

PR #34, opened 2018-05-01
(WIP) (MODULES-6845) Copy V2 Provider into Win32

michaeltlombardi:
Prior to this commit the Win32 provider in this module was a near
mirror of the existing provider in core Puppet.  In order to more
easily maintain the new module, to debug it, and to onboard team
members to maintaining it, this commit copies over the code from
the new provider into the old one. As the new provider was made
to keep in parity with the old one and the tests already existed
to verify this, the change is determined to be low risk.

In a future commit the V2 provider will be updated to manage
scheduled tasks of all compatibility modes and will be the code
which gets all new features and improvements. The legacy provider
will be frozen at a point in the near future for all changes
except for security and critical bug fixes.

This commit also updates the test suite to verify the changes and
removes unneccessary logic from the tests which were needed when
the providers did not share helper code.

However, in this iteration, the tests only run over the legacy
provider and not the V2 provider, though they are using the same
helper at this time.  In a future commit, when the V2 provider is
updated to use the new V2 helper, the test suite will again run
for both.

### DIFF
--- a/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
+++ b/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
@@ -124,7 +124,7 @@ case task_provider
 when :win32_taskscheduler
   concrete_klass = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task
 when :taskscheduler_api2
-  concrete_klass = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task
+  concrete_klass = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task
 end
 
 describe Puppet::Type.type(:scheduled_task).provider(task_provider), :if => Puppet.features.microsoft_windows? do
--- a/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
+++ b/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
@@ -120,17 +120,18 @@ describe Puppet::Type.type(:scheduled_task).provider(:taskscheduler_api2), :if =
   end
 end
 
-# The Win32::TaskScheduler and PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task classes should be
-# API compatible and behave the same way.  What differs is which Windows API is used to query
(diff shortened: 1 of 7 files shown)


spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -120,17 +113,17 @@ def time_component
   end
 end
 
-# The Win32::TaskScheduler and PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task classes should be
+# The Win32::TaskScheduler and PuppetX::PuppetLabs::ScheduledTask::V1 classes should be
 # API compatible and behave the same way.  What differs is which Windows API is used to query
 # and affect the system.  This means for testing, any tests should be the same no matter what
 # provider or concrete class (which the provider uses) is used.
-klass_list = Puppet.features.microsoft_windows? ? [Win32::TaskScheduler, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task] : []
```
Iristyle: I think it might be best to do

```ruby
[PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task]
```

Then on the next commit it will become 

```ruby
[PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2V1Task, PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task]
```
Which will probably be more self-evident with respect to changing the api2 provider to use the other helper.

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -525,9 +513,6 @@ def time_component
           'start_minute' => 21,
           'flags'        => 0,
         })
-        @trigger1 = task_provider == :win32_taskscheduler ?
-          { 'trigger_type' => Win32::TaskScheduler::TASK_EVENT_TRIGGER_AT_LOGON, } :
-          nil
         @mock_task.expects(:trigger).with(1).returns(@trigger1)
```
Iristyle: This should be

```ruby	
@mock_task.expects(:trigger).with(1).returns(nil)
```

michaeltlombardi:
Superceded by #27

PR #35, opened 2018-05-04
(MODULES-7150) Refactor Adapter class state management

Iristyle:
Builds on #27, which should be merged prior

- Old code would insert a dummy trigger into brand new tasks
   created for the first time.

   Since they are always immediately removed, and since there is no
   need to insert a dummy trigger, remove that from the code.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -176,10 +176,7 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
 
   def create
     clear_task
-    @task = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new(
-      resource[:name],
-      PuppetX::PuppetLabs::ScheduledTask::Trigger::V1.default_trigger_for('once')
-    )
+    @task = PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2Task.new(resource[:name])
     self.command = resource[:command]
 
     [:arguments, :working_dir, :enabled, :trigger, :user, :compatibility].each do |prop|
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -35,11 +35,6 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
     @task
   end
 
-  def clear_task
-    @task       = nil
-    @triggers   = nil
-  end
-
   def enabled
     task.flags & PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::TASK_FLAG_DISABLED == 0 ? :true : :false
   end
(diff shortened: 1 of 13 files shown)


lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_task.rb
```diff
@@ -51,7 +51,7 @@ def activate(task_name)
     @definition = TaskScheduler2.task_definition(@task)
     @task_password = nil
 
-    @task
```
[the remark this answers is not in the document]
Iristyle: Only callers here are tests, and they can be modified at will.

lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2_v1task.rb
```diff
@@ -49,7 +49,7 @@ def activate(task_name)
     @definition = TaskScheduler2.task_definition(@task)
     @task_password = nil
 
-    @task
```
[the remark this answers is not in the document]
Iristyle: Only callers here are tests, and they can be modified at will.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -175,7 +170,7 @@ def user=(value)
   end
 
   def create
-    clear_task
+    @triggers   = nil
```
glennsarti: nit: double space.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -14,10 +14,9 @@
 
   def self.instances
     PuppetX::PuppetLabs::ScheduledTask::V2Adapter.tasks.collect do |job_file|
```
glennsarti: Can probably rename the iterator here as `task` as job_file has no meaning in the V2 land

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -11,10 +11,9 @@
 
   def self.instances
     PuppetX::PuppetLabs::ScheduledTask::V1Adapter.tasks.collect do |job_file|
```
glennsarti: Can probably rename the iterator here as task as job_file has no meaning in the V2 land

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -655,11 +655,9 @@ def time_component
 
   describe '.instances' do
     it 'should use the list of .job files to construct the list of scheduled_tasks' do
```
glennsarti: Test name should me modified as `.job` is no longer part of the test

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
Iristyle: Done

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
Iristyle: Done

Iristyle:
Ok @glennsarti / @michaeltlombardi ... I think we're ready to go now.  I think the secret sauce here was 989c01e.

I have a few other improvement ideas, but we can wait until later on those.

glennsarti:
Passed through adhoc.  Merging.

PR #37, opened 2018-05-10
(MODULES-6868) Remove unnecessary library include

michaeltlombardi:
Prior to this commit the taskscheduler2 helper required the puppet
win32 code from core puppet; there are no longer any constants or
other calls reliant on this code so we remove it.

### DIFF
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -1,7 +1,4 @@
-require 'puppet/util'
-
 Puppet::Type.newtype(:scheduled_task) do
-  include Puppet::Util
 
   @doc = "Installs and manages Windows Scheduled Tasks.  All attributes
     except `name`, `command`, and `trigger` are optional; see the description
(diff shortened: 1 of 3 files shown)


Iristyle:
I think this looks good from what I can see... not sure if there any other similar removals to be made in the specs?

Also might want to look at the type definition at https://github.com/michaeltlombardi/puppetlabs-scheduled_task/blob/a34cf6c15d7bcfd2dee57d88e0775b568ed5a63c/lib/puppet/type/scheduled_task.rb#L4

It does an `include Puppet::Util` but I don't know that it's necessary to use that mixin. It may be enough to change the require to `require 'puppet/error'` instead, as that appears to be the only code consumed by the `require 'puppet/util'`. I haven't tested for sure, so reserve the right to be wrong on that one ;0

michaeltlombardi:
@Iristyle Turns out you were correct and the mixin could be removed without any ill effects as far as I could see! Tests passed locally and tooling around with puppet apply locally didn't manage to blow anything up.

PR #38, opened 2018-05-14
(MODULES-7120) Refactor Providers / Adapters

Iristyle:
Things remaining for this PR

- [x] Fix / refactor triggers_same? (remove usage of V1::from_manifest_hash)
- [x] Remove V1::from_manifest_hash from win32_taskscheduler_spec.rb
- [x] Remove `Occurrence` class (requires V1::from_manifest_hash to be removed)
- [x] Remove usage of V1::to_manifest_hash, inlining hashes where applicable
- [x] Move code for V1::to_manifest_hash into v1_adapter_spec.rb which should be the only place it's needed
- [x] Remove V2::V1_TYPE_MAP
- [x] Move classes to V2 namespace or Manifest namespace from V1 namespace - updating all callsites

 - Previously the Adapter (v1Adapter / v2Adapter) code returned V1
   trigger representations, which was then consumed by Providers
   (win32_taskscheduler / taskscheduler_api2) that were then
   responsible for converting to manifest hashes.

   Instead, make the adapters reponsible for producing manifest
   hashes so that providers don't have any understanding of the old
   intermediate V1-esque Trigger structure.

   This is the first step towards removing V1 structures.

PR #41, opened 2018-05-21
(MODULES-6526) Fix win32_taskscheduler compatibility

Iristyle:
Builds on #38 which should be merged first

- Commit 77d2d4b introduced the setting of compatibility on the
   win32_taskscheduler provider.

   However, since it set `compatibility =`, the actual setter method
   for the class instance was never properly called, which would
   result in improper compatibility level being set.

   This may vary by host, but on Windows 2008R2, this defaulted to
   creating new tasks with a compatibility level of 2, instead of the
   desired 1.

 - Running a simple manifest like the following, would result in
   Puppet creating a new task with the wrong compatiblity level. Since
   the provider filters on a compatibility level of 1 or lower,
   subsequent runs would never see the existing task, and would think
   it necessary to create the task. In addition to setting the wrong
   compat level, it completely broke idempotency.

```puppet
scheduled_task { 'doIt-v1':
  command => 'C:\\Windows\\System32\\notepad.exe',
  provider => 'win32_taskscheduler',
  trigger => [{schedule => 'daily', minutes_interval => '30', start_time => '15:20'}],
}
```

 - Add new tests that create tasks with the V1Adapter and verify that
   compat is set properly.

   Modify existing tests as well to verify compat set properly.

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/v1adapter.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/v1adapter.rb
@@ -25,7 +25,7 @@ class V1Adapter
       TaskScheduler2.task_definition(@task)
     @task_password = nil
 
-    compatibility = TaskScheduler2::TASK_COMPATIBILITY_V1
+    self.compatibility = TaskScheduler2::TASK_COMPATIBILITY_V1
     set_account_information('',nil)
   end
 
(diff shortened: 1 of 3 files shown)


glennsarti:
Hrmm odd...Should the acceptance tests be updated as they're supposed to find idempotency issues

https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/master/spec/acceptance/should_create_task_spec.rb#L50-L64

Iristyle:
This has passed adhoc through https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/scheduled_task/job/forge-windows_puppetlabs-scheduled_task_i18n-ruby_adhoc/16/

PR #42, opened 2018-05-21
(MODULES-6267) Implement Boot Trigger

RandomNoun7:
This change implements a boot trigger, allowing tasks to be scheduled to
run when the machine boots.

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -85,7 +85,8 @@ module Trigger
       'once',
       'daily',
       'weekly',
-      'monthly'
+      'monthly',
+      'boot'
      ].freeze
 
     def self.default_trigger_settings_for(schedule = 'once')
@@ -142,8 +143,9 @@ module Trigger
         raise ArgumentError.new("Unknown schedule type: #{manifest_hash["schedule"].inspect}")
       end
 
-      # required fields
-      %w{start_time}.each do |field|
+      required = V2::EVENT_BASED_TRIGGER_MAP.value?(manifest_hash['schedule']) ? [] : %w{start_time}
+
+      required.each do |field|
         next if manifest_hash.key?(field)
         raise ArgumentError.new("Must specify '#{field}' when defining a trigger")
       end
@@ -151,7 +153,7 @@ module Trigger
       start_time_valid = begin Time.parse("2016-5-1 #{manifest_hash['start_time']}"); true rescue false; end
       raise ArgumentError.new("Invalid start_time value: #{manifest_hash['start_time']}") unless start_time_valid
       # The start_time must be canonicalized to match the format that the rest of the code expects
-      manifest_hash['start_time'] = Time.parse(manifest_hash['start_time']).strftime('%H:%M')
+      manifest_hash['start_time'] = Time.parse(manifest_hash['start_time']).strftime('%H:%M') unless manifest_hash['start_time'].nil?
 
       # specific setting rules for schedule types
       case manifest_hash['schedule']
@@ -465,14 +467,26 @@ module Trigger
       TASK_TRIGGER_SESSION_STATE_CHANGE  = 11
     end
 
-    TYPE_MANIFEST_MAP = {
-      Type::TASK_TRIGGER_DAILY => 'daily',
-      Type::TASK_TRIGGER_WEEKLY => 'weekly',
+    SCHEDULE_BASED_TRIGGER_MAP = {
+      Type::TASK_TRIGGER_DAILY      => 'daily',
+      Type::TASK_TRIGGER_WEEKLY     => 'weekly',
       # NOTE: monthly uses context to determine MONTHLY or MONTHLYDOW
-      Type::TASK_TRIGGER_MONTHLY => 'monthly',
+      Type::TASK_TRIGGER_MONTHLY    => 'monthly',
       Type::TASK_TRIGGER_MONTHLYDOW => 'monthly',
-      Type::TASK_TRIGGER_TIME => 'once',
+      Type::TASK_TRIGGER_TIME       => 'once',
     }.freeze
+
+    EVENT_BASED_TRIGGER_MAP = {
+      Type::TASK_TRIGGER_BOOT                 => 'boot',
+      # The triggers below are not yet supported.
+      # Type::TASK_TRIGGER_EVENT                => 'event',
+      # Type::TASK_TRIGGER_IDLE                 => 'idle',
+      # Type::TASK_TRIGGER_REGISTRATION         => 'task_registered',
+      # Type::TASK_TRIGGER_LOGON                => 'logon',
+      # Type::TASK_TRIGGER_SESSION_STATE_CHANGE => 'session_state_change',
+    }.freeze
+
+    TYPE_MANIFEST_MAP = (SCHEDULE_BASED_TRIGGER_MAP.merge(EVENT_BASED_TRIGGER_MAP)).freeze
 
     def self.type_from_manifest_hash(manifest_hash)
       # monthly schedule defaults to TASK_TRIGGER_MONTHLY unless...
@@ -530,6 +544,10 @@ module Trigger
             'which_occurrence' => occurrences.first || '',
             'day_of_week'      => Day.bitmask_to_names(iTrigger.DaysOfWeek),
           })
+        when Type::TASK_TRIGGER_BOOT
+          manifest_hash.merge!({
+            'schedule' => 'boot'
+          })
       end
 
       manifest_hash
@@ -559,7 +577,7 @@ module Trigger
       # when start_date is null or missing, Time.parse returns today
       datetime_string = "#{manifest_hash['start_date']} #{manifest_hash['start_time']}"
       # Time.parse always assumes local time
-      iTrigger.StartBoundary = Time.parse(datetime_string).iso8601
+      iTrigger.StartBoundary = Time.parse(datetime_string).iso8601 unless datetime_string.strip.empty?
 
       # ITrigger specific settings
       case iTrigger.Type
--- a/spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
+++ b/spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
@@ -121,6 +121,23 @@ describe PuppetX::PuppetLabs::ScheduledTask::Trigger::Manifest do
       expect(canonical).to eq(expected)
     end
 
+    [
+      {'schedule' => 'boot'},
+    ].each do |event_manifest|
+
+      describe 'when validating event based triggers' do
+
+        it 'should allow a nil start_time for event based triggers' do
+          expect {subject.class.canonicalize_and_validate(event_manifest)}.to_not raise_error(ArgumentError)
+        end
+
+        it 'should not set a default date' do
+          validated = subject.class.canonicalize_and_validate(event_manifest)
+          expect(validated).not_to have_key('start_date')
+        end
+      end
+    end
+
     it 'should canonicalize `start_time` to %H:%M' do
       manifest_hash = MINIMAL_MANIFEST_HASH.merge({ 'start_time' => '2:03 pm' })
       expected = MINIMAL_MANIFEST_HASH.merge({ 'start_time' => '14:03' })
@@ -911,6 +928,15 @@ describe PuppetX::PuppetLabs::ScheduledTask::Trigger::V2 do
 
     [
       { :ole_type => 'IBootTrigger', :Type => V2::Type::TASK_TRIGGER_BOOT, },
+    ].each do |trigger_details|
+      it "should convert an #{trigger_details[:ole_type]} instance" do
+        # stub is not usable outside of specs (like in DEFAULT_V2_ITRIGGER_PROPERTIES)
+        iTrigger = stub(DEFAULT_V2_ITRIGGER_PROPERTIES.merge(trigger_details))
+        expect { subject.class.to_manifest_hash(iTrigger) }.to_not raise_error(ArgumentError)
+      end
+    end
+
+    [
       { :ole_type => 'IIdleTrigger', :Type => V2::Type::TASK_TRIGGER_IDLE, },
       { :ole_type => 'IRegistrationTrigger', :Type => V2::Type::TASK_TRIGGER_REGISTRATION, },
       { :ole_type => 'ILogonTrigger', :Type => V2::Type::TASK_TRIGGER_LOGON, },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Added
+
+- `boot` trigger support ([MODULES-6267](https://tickets.puppetlabs.com/browse/MODULES-6267))
+
 ## [0.2.0] - 2018-05-09
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -145,6 +145,25 @@ scheduled_task { 'Disk Cleanup Monthly First Saturday':
 }
 ~~~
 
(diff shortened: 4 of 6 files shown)


lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -465,15 +468,27 @@ class Type
...
+      # These are event based triggers that will not require a start time.
+      Type::TASK_TRIGGER_BOOT                 => 'boot',
+      Type::TASK_TRIGGER_IDLE                 => 'idle',
+      Type::TASK_TRIGGER_REGISTRATION         => 'registration',
+      Type::TASK_TRIGGER_LOGON                => 'logon',
+      Type::TASK_TRIGGER_SESSION_STATE_CHANGE => 'session_state_change',
+      Type::TASK_TRIGGER_EVENT                => 'event',
     }.freeze
 
+    TYPE_MANIFEST_MAP = (TIME_TRIGGER_COLLECTION.merge(EVENT_TRIGGER_COLLECTION)).freeze
```
Iristyle: The current collection defines which trigger types are supported, so I don't think we can add anything here yet without some other code being changed. So I think it might be simpler to just add `boot` to the existing collection for now and handle the one-off differences as we are presently, and tackle the `start_date` / `start_time` problem later.

I'm also not sure if this grouping like this for the sake of finding required values is the right way to go given the chart at https://gist.github.com/Iristyle/58ff446bd3f60e68d52df3868c917a0e#trigger-specific-properties

I had given this a little thought earlier, specifically if we have a way of defining things like optional / required values in the hash from the manifest:

```ruby
triggers = {
Type::TASK_TRIGGER_DAILY                => {
  'name' => 'daily',
  'required' => ['start_time'],
  'optional' => ['start_date', 'end_date', 'minutes_interval', 'minutes_duration', 'every'],
},
Type::TASK_TRIGGER_WEEKLY                => {
  'name' => 'weekly',
  'required' => ['start_time'],
  'optional' => ['start_date', 'end_date', 'minutes_interval', 'minutes_duration', 'every', 'day_of_week'],
},
}
```

This would make the code a lot more self-documenting, as we don't have any sort of clear definition of this mapping in code (aside from mentally unwinding the `canonicalize_and_validate` or similar code).

RandomNoun7: Changed handling boot trigger exemptions to handling the specific case instead of Event Triggers generally.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -497,8 +503,8 @@ def self.to_manifest_hash(iTrigger)
         'start_date'       => start_boundary ? start_boundary.strftime('%Y-%-m-%-d') : '',
         'start_time'       => start_boundary ? start_boundary.strftime('%H:%M') : '',
         'enabled'          => iTrigger.Enabled,
-        'minutes_interval' => Duration.to_minutes(iTrigger.Repetition.Interval) || 0,
-        'minutes_duration' => Duration.to_minutes(iTrigger.Repetition.Duration) || 0,
+        'minutes_interval' => iTrigger.Repetition.Interval == '' ? 0 : Duration.to_minutes(iTrigger.Repetition.Interval),
```
Iristyle: I don't think these need to be changed... see https://gist.github.com/Iristyle/58ff446bd3f60e68d52df3868c917a0e#gistcomment-2387093 and `Duration.to_minutes` impl at https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/master/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb#L47-L55

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -555,11 +561,13 @@ def self.append_trigger(definition, manifest_hash)
...
-      # when start_date is null or missing, Time.parse returns today
-      datetime_string = "#{manifest_hash['start_date']} #{manifest_hash['start_time']}"
-      # Time.parse always assumes local time
-      iTrigger.StartBoundary = Time.parse(datetime_string).iso8601
+      if manifest_hash['start_date'] || manifest_hash['start_time']
+        # manifests specify datetime in the local timezone, ITrigger accepts ISO8601
+        # when start_date is null or missing, Time.parse returns today
+        datetime_string = "#{manifest_hash['start_date']} #{manifest_hash['start_time']}"
+        # Time.parse always assumes local time
+        iTrigger.StartBoundary = Time.parse(datetime_string).iso8601
```
Iristyle: Alternatively...

```ruby
iTrigger.StartBoundary = Time.parse(datetime_string).iso8601 unless datetime_string.strip.empty?
```

I'm fine with whichever you think reads better.

```
[3] pry(main)> " ".strip.empty?
=> true
```

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -924,6 +940,18 @@
       end
     end
 
+    [
+      { :ole_type => 'IBootTrigger', :Type => V2::Type::TASK_TRIGGER_BOOT, },
+    ].each do |trigger_details|
+      it "should convert an #{trigger_details[:ole_type]} instance" do
+        # stub is not usable outside of specs (like in DEFAULT_V2_ITRIGGER_PROPERTIES)
+        iTrigger = stub(DEFAULT_V2_ITRIGGER_PROPERTIES.merge(trigger_details))
+        iTrigger.Repetition.stubs(:Interval).returns('')
```
Iristyle: I don't think the stubs here are needed, are they?

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -121,6 +121,23 @@
       expect(canonical).to eq(expected)
     end
 
+    [
+      {'schedule' => 'boot'},
+    ].each do |event_manifest|
+
+      describe 'when validating event based triggers' do
```
Iristyle: I'm wondering if we can use another term than `event`... but I can't think of anything better.

My concerns:

* There's an `IEventTrigger` that means something very specifically (though we'll probably never implement it)
* Arguably you could say that a trigger set for a specific date is responding to a time-based "event"

Maybe just referring to them everywhere as `event based` is sufficient?

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -121,6 +121,23 @@
...
+
+        it 'should allow a nil start_time for event triggers' do
```
Iristyle: Maybe `event based` here?

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
```diff
@@ -121,6 +121,23 @@
...
+          expect {subject.class.canonicalize_and_validate(event_manifest)}.to_not raise_error(ArgumentError)
+        end
+
+        it 'should not set a default date' do
+          validated = subject.class.canonicalize_and_validate(event_manifest)
+          expect(validated.has_key?('start_date')).to eq(false)
```
Iristyle: I think this can use a [predicate matcher](https://relishapp.com/rspec/rspec-expectations/v/2-4/docs/built-in-matchers/predicate-matchers) instead:

```ruby
expect(validated).not_to have_key('start_date')
```

This will produce better output from rspec in a failure

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -142,16 +143,20 @@ def self.canonicalize_and_validate(manifest_hash)
...
 
+      # The boot trigger does not require a start time because it is an event
+      # based trigger, not a schedule. This is handled as a one off for now
+      # until we can make the changes to the module required to handle event
+      # based triggers as a general case.
       # required fields
       %w{start_time}.each do |field|
         next if manifest_hash.key?(field)
-        raise ArgumentError.new("Must specify '#{field}' when defining a trigger")
+        raise ArgumentError.new("Must specify '#{field}' when defining a trigger") unless manifest_hash['schedule'] == 'boot'
```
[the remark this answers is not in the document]
RandomNoun7: I agree that checking for this inside a loop isn't great. In my previous implementation I had checked for the general case of `EVENT_BASED_TRIGGER.value?(manifest_hash['schedule'])`. That check was done in an unless block that prevented this code from running at all if it was an event trigger. That way this would never have to be modified until we ran into more complex requirements than either or, in which case this would likely need to be moved into a function anyway. 

If you don't think that would be better I'll go ahead and make this modification.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
RandomNoun7: Given the prior changes to the way Duration was getting checked these stubs were required. The current mock returns Interval and Duration as hash keys and the way I had changed the code required a mock of a method or the tests broke. Since I have reverted those changes based on comments above these stubs are indeed no longer needed and I am removing them.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
RandomNoun7: I also tried to come up with a term for these kinds of triggers that avoided the word event, but come up empty for something genuinely better.The saving factor is that we can name each type of event we want to support in the `schedule` property with names like `boot` or `logon`, but on the backend call them `event` since their properties are so similar. Personally I think it would be cool to support `IEventTrigger` some day, and we could call that one `event`, but as you say, maybe never get to it.

I'm fine with event based, but given that this describe string already says `event based`, are you asking for a change to this string or just the one below?

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
RandomNoun7: Changing as per above.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
RandomNoun7: Awesome. That does indeed read better.

spec/unit/puppet_x/puppetlabs/scheduled_task/trigger_spec.rb
Iristyle: I think as long as we stick to `event based` for now, its fine.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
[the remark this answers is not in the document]
Iristyle: Yeah, I think the condition is kind of hiding in the loop, which makes for a little bit of a trap if something else is added to the required list with `start_time` (at the very least it would make for multiple code changes instead of just one)... 

The check could also be moved up a line higher to say `next if manifest_hash.key?(field) || (manifest_hash['schedule'] == 'boot')` but I think that suffers the same problem as what we've got now.

I think setting the required fields up-front is clearer in this case.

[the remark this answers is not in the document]
RandomNoun7: To clarify though what I meant was more like 
```
%w{start_time}.each do |field|
  next if manifest_hash.key?(field)
  raise ArgumentError.new("Must specify '#{field}' when defining a trigger")
end unless EVENT_TRIGGER_COLLECTION.value?(manifest_hash['schedule'])
# Is it bad form to put an unless at the end of a block like this?
```
This way the entire loop is skipped if it's a boot trigger, you can add fields to the array as needed, and even if we support new event based triggers in the future we still don't need to edit this check. Adding the full list of event based triggers to an event based triggers collection doesn't make it a free for all for people to put those trigger types in the manifest because we just wouldn't add types we haven't tested to the `ValidScheduleKeys` array.

So yeah, if that kind of check is also a bad idea I'll implement your original suggestion.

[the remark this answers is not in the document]
Iristyle: Yeah, you can do the conditional after the `end` like that, but it's usually considered bad form for readability reasons - see Rubocop rule at https://github.com/bbatsov/ruby-style-guide#no-multiline-if-modifiers

I think we'll probably end up with something like your idea eventually: 

```ruby
required = EVENT_TRIGGER_COLLECTION.value?(manifest_hash['schedule']) ? [] : %w{start_time}
```

But hopefully it's based off the model idea I was tossing around and is more like:

```ruby
required = TRIGGER_DEFINITIONS[manifest_hash['schedule']].required
```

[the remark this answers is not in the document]
RandomNoun7: Well, until we get the model definitions idea worked out, we could move the unless from the end of the block to an enclosing
```
unless <condition>
  <do stuff>
end
```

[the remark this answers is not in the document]
Iristyle: An empty array `[]` for required fields or an `unless` like that are both fine I think... either way.

RandomNoun7:
This is still WIP because I need to do some additional manual testing of trigger properties in conjunction with the `boot` trigger, and I need to add an idempotency test to the spec to ensure that doesn't get regressed since that's the exact bug I ran into during manual testing.

RandomNoun7:
Also I just realized that I would rather not say we support boot triggers without supporting `delay`. So working on that now.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -465,15 +467,27 @@ class Type
...
+    }.freeze
+
+    EVENT_BASED_TRIGGER_MAP = {
+      Type::TASK_TRIGGER_BOOT                 => 'boot',
+      # The triggers below are not yet supported.
+      # Type::TASK_TRIGGER_EVENT                => 'event',
+      # Type::TASK_TRIGGER_IDLE                 => 'idle',
+      # Type::TASK_TRIGGER_REGISTRATION         => 'task_registered',
+      # Type::TASK_TRIGGER_LOGON                => 'logon',
+      # Type::TASK_TRIGGER_SESSION_STATE_CHANGE => 'session_state_change',
```
RandomNoun7: @Iristyle These triggers are commented out so we don't forget them, but if they are not commented there is indeed a bug where a task that has an unsupported trigger type manually applied to it, a puppet run will delete that trigger.

RandomNoun7:
@Iristyle No worries. Changes made.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
Iristyle: And specs didn't catch that problem, right?

That's a little concerning because I know that there are some specs around that at https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/master/spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb#L307-L391

We should probably double check that. They may be too mocked to show the real behavior.

RandomNoun7: Yeah I'm not sure just returning nil for an index as in line 317 is sufficient to catch a trigger that exists but is of an unknown type, given the kind of change that I had made.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -465,15 +467,27 @@ class Type
       TASK_TRIGGER_SESSION_STATE_CHANGE  = 11
     end
 
-    TYPE_MANIFEST_MAP = {
-      Type::TASK_TRIGGER_DAILY => 'daily',
-      Type::TASK_TRIGGER_WEEKLY => 'weekly',
+    TIME_BASED_TRIGGER_MAP = {
```
Iristyle: This should probably be `SCHEDULE` instead of `TIME` ... but we can change later - NBD

PR #49, opened 2018-07-06
(maint) Add all COM error constants from winerror.h

Iristyle:
- There are a number of COM HRESULT values specific to the task
   scheduler APIs.  Namespace all of those values in the module
   PuppetX::PuppetLabs::ScheduledTask::TaskScheduler2::Error
   Add a helper function that can convert values to their 2s
   complement, which is somehow how APIs may return them.

   For instance, for values about 0x800XXXXX, it's common to see both
   usages in the wild.  For instance, ERROR_FILE_NOT_FOUND can either
   be 0x80070002 or -2147024894.

   Typically the WIN32OLERuntimeError class will contain messages
   that are using the 80070002 style format.

 - Rework code that catches WIN32OLERuntimeError, adding a helper
   method to match on the embedded string values for a given HRESULT
   in both formats.

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
@@ -1,6 +1,8 @@
 # The TaskScheduler2 class encapsulates taskscheduler settings and behavior using the v2 API
 # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383600(v=vs.85).aspx
 
+require_relative './error'
+
 # @api private
 module PuppetX
 module PuppetLabs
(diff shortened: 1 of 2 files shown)


lib/puppet_x/puppetlabs/scheduled_task/taskscheduler2.rb
```diff
@@ -71,6 +71,12 @@ module TaskScheduler2
   TASK_FLAG_SYSTEM_REQUIRED              = 0x1000
   TASK_FLAG_RUN_ONLY_IF_LOGGED_ON        = 0x2000
 
+  def self.is_com_error(win32OLERuntimeError, hresult)
+    # does not include 0x prefix
+    hex = (hresult & Error::MAX_32_BIT_VALUE).to_s(16)
```
Iristyle: Seems a little silly to take the original hex value, convert to 2s complement for storage... and then revert to `0x800XXXXX` style for a string comparison.

Maybe better to just store as `0x800XXXXX` and add a helper to `Error` class to use that format when numbers need to be compared.

PR #52, opened 2018-07-07
(MODULES-7425) Remove unnecessary TaskScheduler2 methods

Iristyle:
~~Builds on #51, which should be merged first~~

Removes a number of code paths from `TaskScheduler2` that were unnecessary and/or unnecessarily complex. Many of these methods took a `definition` object and only performed a single action with it, and could easily be inlined into the two adapter classes.

This is the first part of a class restructuring that will:

* Merge the 2 adapter classes into a single base implementation (since they share 99% of their impl)
* Potentially refactor remaining methods from `TaskScheduler2` directly into the base adapter class (there's only about 120 lines of actual code in `TaskScheduler2` module now - and it would probably be fine)
* Ultimately this should simplify the class hierarchy so that maintenance is simpler

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/error.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/error.rb
@@ -11,6 +11,14 @@ module TaskScheduler2
       -(-hresult & MAX_32_BIT_VALUE)
     end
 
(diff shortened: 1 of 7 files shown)


michaeltlombardi:
Passed adhoc, ready for merge!

PR #54, opened 2018-07-13
(MODULES-7463) Refactor V1Adapter and TaskScheduler2 into one class

Iristyle:
Builds on #53 which must be merged first.


michaeltlombardi:
Passed adhoc!

PR #55, opened 2018-07-31
(MAINT) Add legacy taskscheduler code as helper and freeze

michaeltlombardi:
Prior to this commit the spec tests relied on the built-in
utility code for the windows task scheduler inside of core
puppet. With the Puppet 6 release this code will be removed,
meaning the tests relying on this reference will fail.

This commit copies that code into a helper file in the spec
folder and ensures the tests relatively require it instead
of pointing to the utility baked into Puppet.

This should ensure the tests continue to pass and backwards
compatibility can be safely maintained once Puppet 6 is
released.

### DIFF
--- a/spec/integration/puppet_x/puppetlabs/scheduled_task/v1adapter_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/scheduled_task/v1adapter_spec.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 require 'spec_helper'
 
-require 'puppet/util/windows/taskscheduler' if Puppet.features.microsoft_windows?
+require_relative '../../../../legacy_taskscheduler' if Puppet.features.microsoft_windows?
 require 'puppet_x/puppetlabs/scheduled_task/v1adapter'
 
 # These integration tests confirm that the tasks created in a V1 scheduled task APi are visible
--- a/spec/legacy_taskscheduler.rb
+++ b/spec/legacy_taskscheduler.rb
@@ -1,0 +1,1272 @@
+# DO NOT EDIT, ALTER, OR DELETE THIS FILE
+# This file is copied from core puppet and frozen
+# It is used in spec tests to ensure compatibility
+# between this module and the old built-in type.
+# Frozen from: https://github.com/puppetlabs/puppet/blob/5.5.3/lib/puppet/util/windows/taskscheduler.rb
+require 'puppet/util/windows'
+
+# The Win32 module serves as a namespace only
+module Win32
+  # The TaskScheduler class encapsulates taskscheduler settings and behavior
+  class TaskScheduler
+    include Puppet::Util::Windows::String
+
+    require 'ffi'
+    extend FFI::Library
+
+    # The error class raised if any task scheduler specific calls fail.
+    class Error < Puppet::Util::Windows::Error; end
+
+    class << self
+      attr_accessor :com_initialized
+    end
+
+    # :stopdoc:
+    TASK_TIME_TRIGGER_ONCE            = :TASK_TIME_TRIGGER_ONCE
+    TASK_TIME_TRIGGER_DAILY           = :TASK_TIME_TRIGGER_DAILY
+    TASK_TIME_TRIGGER_WEEKLY          = :TASK_TIME_TRIGGER_WEEKLY
+    TASK_TIME_TRIGGER_MONTHLYDATE     = :TASK_TIME_TRIGGER_MONTHLYDATE
+    TASK_TIME_TRIGGER_MONTHLYDOW      = :TASK_TIME_TRIGGER_MONTHLYDOW
+    TASK_EVENT_TRIGGER_ON_IDLE        = :TASK_EVENT_TRIGGER_ON_IDLE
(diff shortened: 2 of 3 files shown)


spec/legacy_taskscheduler.rb
```diff
@@ -0,0 +1,1271 @@
+# DO NOT EDIT, ALTER, OR DELETE THIS FILE
```
glennsarti: Probably want to reference which commit this file came from e.g.

https://github.com/puppetlabs/puppet/blob/5.5.3/lib/puppet/util/windows/taskscheduler.rb

spec/legacy_taskscheduler.rb
```diff
@@ -0,0 +1,1271 @@
...
+               :MinutesInterval, :dword,         # Run bracket repetition interval.
+               :rgFlags, :dword,                 # Trigger flags.
+               :TriggerType, TASK_TRIGGER_TYPE,  # Trigger type.
+               :Type, TRIGGER_TYPE_UNION,        # Trigger data.
+               :Reserved2, :word,                # Reserved. Must be zero.
+               :wRandomMinutesInterval, :word    # Maximum number of random minutes after start time
+      end
+    end
+  end
+end
```
glennsarti: missing newline

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
-require 'puppet/util/windows/taskscheduler' if Puppet.features.microsoft_windows?
+require_relative '../../../../legacy_taskscheduler' if Puppet.features.microsoft_windows?
```
glennsarti: Should this be in the `spec_helper.rb` and then removed from individual test files? or not worth it?

michaeltlombardi: I'm thinking _possibly_ not since it's only needed for two test files, not the whole suite?

spec/legacy_taskscheduler.rb
michaeltlombardi: Fixed,

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
Iristyle: I prefer it to be in the two files explicitly where it's needed (as it is now), rather than pollute all the other tests.

spec/integration/puppet_x/puppetlabs/scheduled_task/v1adapter_spec.rb
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 require 'spec_helper'
 
-require 'puppet/util/windows/taskscheduler' if Puppet.features.microsoft_windows?
+require_relative '../../../../legacy_taskscheduler' if Puppet.features.microsoft_windows?
```
Iristyle: One other thing I should have mentioned earlier.

I think we should do one of two things here:

* Guard the require additionally to only load where `!defined?(Win32::Taskscheduler)`
    * downside is unlikely divergent behavior if someone accidentally modifies the Puppet 4 / Puppet 5 code
    * upside is it prevents a number of duplicate redefinition warnings for the constants
* Include some notes that:
    * we're pulling in deprecated code
    * the code being pulled in will override / reopen classes already defined in the Puppet source / gem for versions < 6

I'd prefer the former in this case. Thoughts?

Iristyle: The defensive guard was me thinking that somewhere else code had already called `require 'puppet/util/windows/taskscheduler'` (indirectly through `require 'puppet/util/windows'` for instance). But looking at Puppet, that may not be the case. Worth a quick check to see if the guard is unnecessary.

michaeltlombardi: When I run the spec tests locally I _do_ get the warnings about the constants. Going to update with the guard.

michaeltlombardi: It looks like rspec has loaded `RSpec::Puppet::Win32::TaskScheduler` at some point before this requires and causes tests to fail - it's also the reason we're seeing the warnings about constants, I think. Investigating...

michaeltlombardi: Aha! It's being loaded via `rspec-puppet`:

Iristyle: You could look for `Win32::TaskScheduler::COM` instead, which isn't defined in rspec-puppet.

joshcooper: Good god, why is rspec-puppet doing that...

michaeltlombardi: I don't think it's ever the real thing, at this point - and the updated guard doesn't prevent the reloading of those constants.

PR #56, opened 2018-08-10
(maint) Refactor trigger management on task

Iristyle:
- Both existing providers do the same thing when iterating triggers
   on a task. They retrieve a count, enumerate each trigger, removing
   unsupported types, all the while attaching the index to the
   converted hash.

   Since this code is duplicated, and since the Task class already
   implements the retrieval / conversion to a manifest_hash on
   individual tasks, it made sense to move this duplicated code into
   a single #triggers method on the Task object.

   Note that the index is now returned inside the given manifest hash,
   rather than being appended by the provider code.

 - To keep inline with the existing #trigger(index) code, return nils
   in the returned array for unsupported trigger types. Since the
   Task class is currently designed around ignoring these triggers,
   stay consistent.

 - This allows for the removal of the trigger_count method completely.
   The providers were interested in getting a collection of all
   triggers, not accessing individual ones. There were a couple of
   usages within tests, but these are easily replaced with

 - Also as a result of tests, introduce a new wrapper around task
   clearing, called #clear_triggers. One section of test code
   enumerated tasks, deleting them one by one (using the trigger_count)
   This is unnecessary given this wrapper.

 - Retrieving a trigger at an index can therefore be made private
   since the only caller is in the Task class itself.

 - Update test mocks based on this new setup


spec/integration/puppet_x/puppetlabs/scheduled_task/task_spec.rb
```diff
@@ -275,8 +278,8 @@ def create_task(task_name = nil, task_compatiblity = nil, triggers = [])
         task = subject.new(@task_name)
 
         # trigger specific validation
-        expect(task.trigger_count).to eq(1)
-        expect(task.trigger(0)).to eq(manifest_hash)
+        expect(task.triggers.count).to eq(1)
```
Iristyle: I think this can actually be `match_array`. Will update.

michaeltlombardi:
The second commit's message seems to have been accidentally truncated mid sentence?

PR #57, opened 2018-08-13
(MODULES-7659) PDK Update module due to Beaker 4.0

glennsarti:
This commit updates the module to use PDK 1.6 and the new version of the PDK
template.  This also pins Beaker to 3.0 to stop the current acceptance tests failing until the PDK
template can be changed for Beaker 4.0 compatibility.


glennsarti:
Note that the adhoc git sha's are different as I noticed that some files were CRLF. I fixed them, but the commit changed.

glennsarti:
Passed adhoc.

PR #58, opened 2018-08-13
(MODULES-7129) Add logon trigger

jpogran:
This commit adds the logon event trigger to the module.

### DIFF
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -155,7 +155,15 @@ Puppet::Type.newtype(:scheduled_task) do
           * `which_occurrence` **(Required)** --- The occurrence of the chosen weekday
             when the task should run. Must be one of `first`, `second`, `third`,
             `fourth`, or `fifth`.
-
+      * For `logon` triggers:
+          * `user_id` --- The `user_id` specifies _which_ user this task will trigger
+            for when they logon. If unspecified, or if specified as `undef` or an empty
+            string, the task will trigger whenever **any** user logs on. This property
+            can be specified in one of the following formats:
+            * Local User: `"Administrator"`
+            * Domain User: `"MyDomain\\MyUser"`
+            * SID: `"S-15-..."`
+            * Any User: `''` or `undef`
 
       Examples:
 
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -78,7 +78,8 @@ module Trigger
        'which_occurrence',
        'day_of_week',
        'minutes_interval',
-       'minutes_duration'
+       'minutes_duration',
+       'user_id',
      ].freeze
 
      ValidScheduleKeys = [
@@ -86,7 +87,8 @@ module Trigger
       'daily',
       'weekly',
       'monthly',
-      'boot'
+      'boot',
+      'logon',
      ].freeze
 
     # https://msdn.microsoft.com/en-us/library/system.datetime.fromoadate(v=vs.110).aspx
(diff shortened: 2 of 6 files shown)


lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -636,6 +645,9 @@ def self.append_trigger(definition, manifest_hash)
           iTrigger.MonthsOfYear = Month.indexes_to_bitmask(manifest_hash['months'] || Month.indexes)
           # HACK: convert V1 week value to names, then back to V2 bitmask
           iTrigger.WeeksOfMonth = WeeksOfMonth.names_to_bitmask(manifest_hash['which_occurrence'])
+
+        when Type::TASK_TRIGGER_LOGON
+          iTrigger.UserdId = manifest_hash['user_id']
```
Iristyle: `UserId`, not `UserdId`

Iristyle:
Given manifest

```puppet
scheduled_task { 'logonvalidationfail':
  compatibility => 2,
  command => 'C:\\Windows\\System32\\notepad.exe',
  trigger => [{schedule => 'logon', minutes_interval => '30', start_time => '15:20'}],
}
```

Validation is working 

```
Error: Parameter trigger failed on Scheduled_task[logonvalidationfail]: Must specify 'user_id' when defining a logon trigger at C:/source/puppetlabs-scheduled_task/logon.pp:1
```

This manifest also worked

```puppet
scheduled_task { 'goodlogon':
  compatibility => 2,
  command => 'C:\\Windows\\System32\\notepad.exe',
  trigger => [{schedule => 'logon', minutes_interval => '30', start_time => '15:20', user_id => 'Bob'}],
}

scheduled_task { 'badlogon_compat':
  compatibility => 1,
  command => 'C:\\Windows\\System32\\notepad.exe',
  trigger => [{schedule => 'logon', minutes_interval => '30', start_time => '15:20', user_id => 'Bob'}],
}

scheduled_task { 'badlogon_nousert':
  compatibility => 2,
  command => 'C:\\Windows\\System32\\notepad.exe',
  trigger => [{schedule => 'logon', minutes_interval => '30', start_time => '15:20', user_id => 'JimBob'}],
}
```

```
Notice: Compiled catalog for vagrant-2008r2.localdomain in environment production in 0.15 seconds
Notice: /Stage[main]/Main/Scheduled_task[goodlogon]/ensure: created
Notice: /Stage[main]/Main/Scheduled_task[badlogon_compat]/ensure: created
Error: /Stage[main]/Main/Scheduled_task[badlogon_compat]: Could not evaluate: (in OLE method `RegisterTaskDefinition': )
    OLE error code:80041316 in <Unknown>
      (13,9):UserId:
    HRESULT error code:0x80020009
      Exception occurred.
Notice: /Stage[main]/Main/Scheduled_task[badlogon_nousert]/ensure: created
Notice: Applied catalog in 0.45 seconds
```

Interestingly though, for the one with a non-existent user, the task gets created as `SYSTEM`

lib/puppet/type/scheduled_task.rb
```diff
@@ -155,7 +155,8 @@ def insync?(current)
           * `which_occurrence` **(Required)** --- The occurrence of the chosen weekday
             when the task should run. Must be one of `first`, `second`, `third`,
             `fourth`, or `fifth`.
-
+      * For `logon` triggers:
+          * `user_id` --- The user account that is used to run the task at logon.
```
michaeltlombardi: Is it worth specifying that this can/can't be local or AD accounts?

michaeltlombardi:
@Iristyle I _think_ the reason it doesn't care about the nonexistant user is that it expects to be able to function with AD users, right? So it's not going to check the local machine for users and error out if they're not found.

lib/puppet/type/scheduled_task.rb
michaeltlombardi: Also, is it worth adding the readme example into the type docs, too?

Iristyle:
@michaeltlombardi / @jpogran sorry if my comment about the user wasn't initially clear. If Windows doesn't validate it on creation, there's nothing we can do anyhow... so I think it's totally fine. I was just noting that that was interesting.

lib/puppet/type/scheduled_task.rb
Iristyle: Yeah, I think it would be useful to be a bit more specific, like the `ILogonTrigger` docs.

https://docs.microsoft.com/en-us/windows/desktop/taskschd/logontrigger-userid#property-value

```
The identifier of the user. For example, "MyDomain\MyName" or for a local account, "Administrator".

This property can be in one of the following formats:

User name or SID: The task is started when the user logs on to the computer.
NULL: The task is started when any user logs on to the computer.

Remarks

If you want a task to be triggered when any member of a group logs on to the computer rather than when a specific user logs on, then do not assign a value to the LogonTrigger.UserId property. Instead, create a logon trigger with an empty LogonTrigger.UserId property and assign a value to the principal for the task using the Principal.GroupId property.
```

Reading those docs I don't think we tested `NULL` either.

michaeltlombardi:
@Iristyle, @jpogran  I was unable to replicate your findings where a nonexistant user account could be specified successfully; on my machine this always resulted in an OLE error. Looking into it, the error was around being unable to resolve the specified user.

The updated code and validation ensures a user can be resolved to a SID, much like the user parameter for the overall task.

I've tested this locally and confirmed:

- You can specify a local user who exists
- You can specify a builtin user
- You can specify an empty string or `undef`

Spec all passes, going to kick off against adhoc.

michaeltlombardi:
Per @glennsarti, updated docs and spec tests for the logon trigger.

glennsarti:
@jpogran If you're happy with this, @michaeltlombardi fell free to smash the merge button.

jpogran:
LGTM

PR #59, opened 2018-08-13
(MODULES-7651) Remove spec folder from builds

Iristyle:
- The PDK looks at .pdkignore to determine what to include / exclude
   with the built tarball - see

   puppetlabs/pdk:lib/pdk/module/build.rb@c6db211#L176-L182

 - PMT (Puppet module tool) looks at .pmtignore for the same thing,
   which is how a module is typically bundled in CI or through a
   local workflow with `bundle exec rake build`

### DIFF
--- a/.pdkignore
+++ b/.pdkignore
@@ -15,6 +15,7 @@
 /junit/
 /log/
 /pkg/
+/spec/
 /spec/fixtures/manifests/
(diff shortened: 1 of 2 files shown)


Iristyle:
I verified a local execution of `bundle exec rake build` now removes the `spec` folder.

PR #61, opened 2018-08-14
(maint) Memoize the Schedule.Service COM instance

Iristyle:
- Currently, a new Service instance is created each time:

   * The task method is called to load a task, which is in turn
     called by the initializer for Task
   * For each depth of the folder hierarchy for enum_task_names,
     which is called when retrieving all tasks
   * When a task is deleted
   * When a task is saved

 - This generates a lot of unnecessary object creation overhead and
   can be minimized by memoizing the first created instance

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/task.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/task.rb
@@ -354,10 +354,12 @@ class Task
   private
   # :stopdoc:
   def self.task_service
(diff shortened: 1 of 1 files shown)


Iristyle:
This may lead to fewer intermittent failures... or more?

glennsarti:
Will run through adhoc and then merge.

glennsarti:
Passed smoke testing.  Merging.

PR #65, opened 2018-08-21
Ticket/master/pup 6

melissa:


melissa:
I'll also be coming through with a PR to ci-job-configs to enable localization for realz there so that this file gets updated regularly

PR #73, opened 2019-01-03
(maint) Update pdk template

jpogran:


appveyor.yml
```diff
@@ -1,55 +1,59 @@
...
+  - 'mkdir C:\ProgramData\PuppetLabs\facter && exit 0'
+  - 'mkdir C:\ProgramData\PuppetLabs\hiera && exit 0'
+  - 'mkdir C:\ProgramData\PuppetLabs\puppet\var && exit 0'
+environment:
+  matrix:
+    -
+      RUBY_VERSION: 24-x64
+      CHECK: syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop
+
+    # Warning PDK Convert keeps adding the default matrix cells in instead of replacing them
```
Iristyle: I think the hand editing bit is implicit / easy to overlook - can we make it more explicit? (Also making `Warning` into `WARN:` is a bit more inline with something that requires action IMHO)

PR #76, opened 2019-01-25
pdksync - MODULES-8532 - pdksync_heads/master-0-g7281db5

ThoughtCrhyme:
MODULES-8532 - pdksync_heads/master-0-g7281db5
pdk version: `1.8.0`


RandomNoun7:
@ThoughtCrhyme I'm not sure this one came out right. This update should have removed the Puppet 4 gem from appveyor.yaml so we can stop testing against, I still see it there in diff. I'm wondering if the sync needs to get run again or something?

RandomNoun7:
So the answer here is the Puppet 4 is still in the test matrix because it's still in the test matrix in the sync yaml. This is because the sync yaml has the gems we want to test because this module doesn't support the parallel spec testing pattern that the rest of the modules use. 

I chose to leave Puppet 4 in here and leave the addition of 6 in place. In my opinion we should ticket the work to modify the sync yaml to remove Puppet 4 and add Puppet 6, because that would seem to be the new test matrix we want to test against. That work however is separate from the simple task of updating the pdk template controlled files.

PR #80, opened 2019-03-19
(WIN-280) add skip() unless pattern to tests

ThoughtCrhyme:
to facilitate more complete test output from our unit tests.


ThoughtCrhyme:
I'm not particularly fond of how "bespoke" it felt to add the skip unless pattern to the tests here, especially in task_spec.rb

Feels like the tests could have been structured better to facilitate the new pattern.

PR #81, opened 2019-04-20
(MODULES-7889) convert acceptance tests to Litmus

ThoughtCrhyme:


glennsarti:
Note - The removal of Puppet 4 from spec tests does not mean removing support from the metadata.

PR #83, opened 2019-05-20
(MODULES-7203) Support nonroot task folders

michaeltlombardi:
Prior to this commit the type and providers did not support specifying subfolders in which
to place a scheduled task. This commit adds support to the type and the taskscheduler_api2
provider for specifying scheduled tasks in subfolders.

This feature only exists for tasks whose compatibility is 2 or higher which prevents it from
being used on the legacy win32_taskscheduler provider altogether and necessitates some
validation prior to runtime.

This commit adds such validation by failing with descriptive errors during resource validation.

This new functionality requires the ability to scaffold folders in which to place the scheduled
task, which now happens automatically and only if the folder path does not already exist.

This commit does NOT add functionality for pruning the folders on task deletion, this should be
addressed in a future commit.

It does update the documentation and tests for the new feature as well as ensure all existing
tests continue to pass.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -207,4 +207,10 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
 
     true
   end
+
+  def validate_name
+    if @resource[:name].match?(/\\/) && @resource[:compatibility] < 2
+      raise Puppet::ResourceError, "#{@resource[:name]} specifies a path including subfolders and a compatibility of #{@resource[:compatibility]} - tasks in subfolders are only supported on version 2 and later of the API. Specify a compatibility of 2 or higher or do not specify a subfolder path."
+    end
+  end
 end
--- a/lib/puppet_x/puppetlabs/scheduled_task/task.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/task.rb
@@ -141,9 +141,9 @@ class Task
   #
   def self.tasks(compatibility = V2_COMPATIBILITY)
     enum_task_names(ROOT_FOLDER,
-      include_child_folders: false,
+      include_child_folders: true,
       include_compatibility: compatibility).map do |item|
-        task_name_from_task_path(item)
+        item.partition('\\')[2]
     end
   end
 
@@ -181,9 +181,16 @@ class Task
   end
 
   # Returns whether or not the scheduled task exists.
-  def self.exists?(job_name)
-    # task name comparison is case insensitive
-    tasks.any? { |name| name.casecmp(job_name) == 0 }
+  def self.exists?(task_path)
+    raise TypeError unless task_path.is_a?(String)
+    begin
+      task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+      # https://msdn.microsoft.com/en-us/library/windows/desktop/aa381363(v=vs.85).aspx
+      _task = task_folder.GetTask(task_name_from_task_path(task_path))
+    rescue 
+      return false
+    end
+    true
   end
 
   # Delete the specified task name.
@@ -201,6 +208,7 @@ class Task
   def save
     task_path = @task ? @task.Path : @full_task_path
 
+    self.class.create_folder(self.class.folder_path_from_task_path(task_path))
     task_folder = self.class.task_service.GetFolder(self.class.folder_path_from_task_path(task_path))
     task_user = nil
     task_password = nil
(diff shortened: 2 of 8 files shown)


lib/puppet_x/puppetlabs/scheduled_task/task.rb
```diff
@@ -190,8 +190,21 @@ def self.exists?(job_name)
...
+    delete_empty_folder(folder_path)
+  end
+
+  # Delete empty folders
+  def self.delete_empty_folder(folder_path)
+    parent_folder_path = folder_path_from_task_path(folder_path)
+    if enum_task_names(folder_path, include_child_folders: true).empty?
+      task_service.GetFolder(ROOT_FOLDER).DeleteFolder(folder_path, 0)
+      if enum_task_names(parent_folder_path, include_child_folders: true).empty?
+        delete_empty_folder(parent_folder_path)
```
glennsarti: TODO: Typically recursive calls need a terminating clause e.g. "If parent == nil then exit"

lib/puppet_x/puppetlabs/scheduled_task/task.rb
```diff
@@ -190,8 +190,21 @@ def self.exists?(job_name)
...
-    task_folder = task_service.GetFolder(folder_path_from_task_path(task_path))
+    folder_path = folder_path_from_task_path(task_path)
+    task_folder = task_service.GetFolder(folder_path)
     task_folder.DeleteTask(task_name_from_task_path(task_path), 0)
+    delete_empty_folder(folder_path)
+  end
+
+  # Delete empty folders
+  def self.delete_empty_folder(folder_path)
+    parent_folder_path = folder_path_from_task_path(folder_path)
```
glennsarti: `parent_folder_path` isn't used till later.  may be best to enumerate when it's needed, not before

lib/puppet_x/puppetlabs/scheduled_task/task.rb
```diff
@@ -372,6 +386,18 @@ def self.folder_path_from_task_path(task_path)
     path.empty? ? ROOT_FOLDER : path
   end
 
+  def self.create_folder(path)
+    service = task_service
+    begin
+      service.GetFolder(path)
```
glennsarti: For ref : There is no API call for `exists`.  Just call get and wait for the error.

https://docs.microsoft.com/en-us/windows/desktop/api/taskschd/nn-taskschd-itaskfolder

lib/puppet_x/puppetlabs/scheduled_task/task.rb
```diff
@@ -372,6 +386,18 @@ def self.folder_path_from_task_path(task_path)
...
+    rescue WIN32OLERuntimeError => e
+      unless Error.is_com_error_type(e, Error::ERROR_FILE_NOT_FOUND)
+        raise Puppet::Error.new( _("GetTask failed with: %{error}") % { error: e }, e )
+      end
+      service.GetFolder("\\").CreateFolder(path)
```
glennsarti: Do we want to trap errors for this?

lib/puppet_x/puppetlabs/scheduled_task/task.rb
```diff
@@ -372,6 +386,18 @@ def self.folder_path_from_task_path(task_path)
     path.empty? ? ROOT_FOLDER : path
   end
 
+  def self.create_folder(path)
+    service = task_service
+    begin
+      service.GetFolder(path)
+    rescue WIN32OLERuntimeError => e
+      unless Error.is_com_error_type(e, Error::ERROR_FILE_NOT_FOUND)
+        raise Puppet::Error.new( _("GetTask failed with: %{error}") % { error: e }, e )
```
glennsarti: typo: Should be `GetFolder`

lib/puppet_x/puppetlabs/scheduled_task/task.rb
michaeltlombardi: :sob:

lib/puppet_x/puppetlabs/scheduled_task/task.rb
michaeltlombardi: 😭

lib/puppet_x/puppetlabs/scheduled_task/task.rb
glennsarti: Probably not

michaeltlombardi: I'm leaning towards "no" initially, but I'm open to it.

michaeltlombardi:
This is working its way through Jenkins now. :)

michaeltlombardi:
Passed adhoc!

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -207,4 +207,10 @@ def validate_trigger(value)
 
     true
   end
+
+  def validate_name
+    if @resource[:name].match?(/\\/) && @resource[:compatibility] < 2
+      raise Puppet::ResourceError, "#{@resource[:name]} specifies a path including subfolders and a compatibility of #{@resource[:compatibility]} - tasks in subfolders are only supported version 2 and later of the API. Specify a compatibility of 2 or higher or do not specify a subfolder path."
```
RandomNoun7: `tasks in subfolders are only supported version 2 and` Consider adding an `in` or an `on` between supported and version.

PR #86, opened 2019-07-19
(MODULES-9370) Align module to core structure

florindragos:


spec/spec_helper_acceptance.rb
```diff
@@ -6,7 +6,7 @@
 
 def add_test_user
```
tphoney: this does not seem to follow the litmus template for spec_helper_acceptance. These helpers should be in spec_helper_acceptance_local.rb

spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -20,1586 +20,1553 @@
...
-          },
-          {
-            'start_date'       => '2013-12-15',
-            'start_time'       => '16:23',
-            'schedule'         => 'once',
-            'minutes_interval' => 60,
-            'minutes_duration' => 240,
-            'enabled'          => true,
-            'index'            => 2,
-          }
```
tphoney: this block of tests, seems to have dissappeared

florindragos: it is identical to the one below

spec/spec_helper_acceptance.rb
ThoughtCrhyme: It is consistent with the documentation https://github.com/puppetlabs/puppet_litmus/wiki/Converting-a-module-to-use-Litmus#specspec_helper_acceptancerb

ThoughtCrhyme: Also, if I may ask, why are we adding "local" to the end of the file? That seems to imply that it's only used for localhost acceptance testing which isn't the case.

tphoney: @ThoughtCrhyme If you look at any of the modules, the _local.rb files denote local differences to the templated ruby files. hence https://github.com/puppetlabs/puppetlabs-apt/blob/master/spec/spec_helper_local.rb https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/spec_helper_local.rb for unit tests. https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/spec_helper_acceptance_local.rb and https://github.com/puppetlabs/puppetlabs-apt/blob/master/spec/spec_helper_acceptance_local.rb for non standard functions. If you feel this is unclear please update the wiki. 

The wiki states to use the template provided.

michaeltlombardi:
Reworked and merged in #88

PR #88, opened 2019-08-01
(MODULES-9370) Module Alignment

michaeltlombardi:
Building on the work by @florindragos in #86, this PR makes a few small changes to the actual files but largely just splits the commits to ensure only the tests or library code are being changed at a time to keep confidence that no regressions are introduced.


.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
```
ThoughtCrhyme: Are we tracking vscode recommendations per module?

appveyor.yml
```diff
@@ -73,4 +73,4 @@ notifications:
       - nobody@nowhere.com
     on_build_success: false
     on_build_failure: false
-    on_build_status_changed: false
+    on_build_status_changed: false
```
ThoughtCrhyme: line ending

metadata.json
```diff
@@ -30,10 +30,10 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.9.0 < 7.0.0"
+      "version_requirement": ">= 5.5.10 < 7.0.0"
```
ThoughtCrhyme: Be sure to mark this PR as backwards incompatible since we're doing a major version bump here.

lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
```diff
@@ -143,9 +142,9 @@ def trigger=(value)
   end
 
   def user=(value)
-    self.fail("Invalid user: #{value}") unless Puppet::Util::Windows::SID.name_to_sid(value)
+    raise("Invalid user: #{value}") unless Puppet::Util::Windows::SID.name_to_sid(value)
 
-    if value.to_s.downcase != 'system'
+    if !value.to_s.casecmp('system').zero?
```
ThoughtCrhyme: Not a blocker, but, add this to my growing list of rubocop suggestions that I believe are more obtuse than the original code.

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -1,674 +1,678 @@
...
-          %w{day_of_week which_occurrence}.each do |field|
-            next if manifest_hash.key?(field)
-            raise ArgumentError.new("#{field} must be specified when creating a monthly day-of-week based trigger")
+        required.each do |field|
+          next if manifest_hash.key?(field)
+          raise ArgumentError, "Must specify '#{field}' when defining a trigger"
+        end
+
+        start_time_valid = time_valid?(manifest_hash['start_time'])
+        raise ArgumentError, "Invalid start_time value: #{manifest_hash['start_time']}" unless start_time_valid
```
ThoughtCrhyme: I don't understand why these errors don't need to be decorated like _("Invalid start_time value...")

.vscode/extensions.json
michaeltlombardi: I think this is PDK-ified now?

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
michaeltlombardi: TBH, neither do I? But rubocop isn't yelling in this case.

kinda confused

PR #89, opened 2019-08-08
(MODULES-9382) Stringify module

florindragos:


REFERENCE.md
```diff
@@ -0,0 +1,223 @@
+# Reference
+<!-- DO NOT EDIT: This document was generated by Puppet Strings -->
+
+## Table of Contents
+
+**Resource types**
+
+* [`scheduled_task`](#scheduled_task): Installs and manages Windows Scheduled Tasks.  All attributes except `name`, `command`, and `trigger` are optional; see the description of th
```
RandomNoun7: This sentence gets cut off at the end.

RandomNoun7: It looks like there may be a workaround by using a `@summary` tag.
https://tickets.puppetlabs.com/browse/MODULES-8183

florindragos: Looks like there is a character limit for `@summary` so I used a combination of `@summary` and `@note`

PR #119, opened 2020-02-06
(MODULES-10101) Use RunOnLastWeekOfMonth for which_occurrence = last

cmccrisken-puppet:
### Description
We were not correctly determining the idempotency of a scheduled task that had a [MonthlyDOWTrigger](https://docs.microsoft.com/en-gb/windows/win32/taskschd/monthlydowtrigger) set to `last`:
```
'which_occurrence' => 'last'
```
The task was being correctly set on the Windows Server instances, but when we read back the task, we were determining that the `Task.Definition.Triggers[i].WeeksOfMonth` attribute was `0` (i.e. nothing defined). Instead, we should be checking the [RunOnLastWeekOfMonth](https://docs.microsoft.com/en-gb/windows/win32/taskschd/monthlydowtrigger-runonlastweekofmonth) to determine if the task is set to run on the last week of the month.

**NOTE:** We are not changing how this parameter is configured, just how it is read back.
### Tests
- [x] `pdk validate` successfully passes
- [x] Existing acceptance tests pass on all compatible OSs
- [x] New acceptance tests pass on all compatible OSs

### DIFF
--- a/spec/acceptance/should_create_task_spec.rb
+++ b/spec/acceptance/should_create_task_spec.rb
@@ -220,4 +220,35 @@ describe 'Should create a scheduled task' do
       expect(result.stdout).to match(%r{#{username2}})
     end
   end
+
+  it "correctly determines idempotency for tasks with LastWeekOfMonth='last'" do
+    pp = <<-MANIFEST
+    scheduled_task {'#{taskname}':
+      ensure        => present,
+      command       => 'c:\\\\windows\\\\system32\\\\notepad.exe',
+      arguments     => "foo bar baz",
+      working_dir   => 'c:\\\\windows',
+      user          => '#{username}',
+      password      => '#{password}',
+      trigger   => {
+        'start_date'       => '2019-9-01',
+        'start_time'       => '05:40',
+        'minutes_interval' => '0',
+        'minutes_duration' => '0',
+        'schedule'         => 'monthly',
+        'months'           => [1,2,3,4,5,6,7,8,9,10,11,12],
+        'which_occurrence' => 'last',
+        'day_of_week'      => ['tues']
+      },
+    }
+    MANIFEST
+    idempotent_apply(pp)
+
+    # Verify the task exists
+    query_cmd = "schtasks.exe /query /v /fo list /tn #{taskname}"
+
+    run_shell(query_cmd) do |result|
+      expect(result.stdout).to match(%r{Last\sTUE})
+    end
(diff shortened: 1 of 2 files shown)


spec/acceptance/should_create_task_spec.rb
```diff
@@ -220,4 +220,35 @@
...
+      },
+    }
+    MANIFEST
+    idempotent_apply(pp)
+
+    # Verify the task exists
+    query_cmd = "schtasks.exe /query /v /fo list /tn #{taskname}"
+
+    run_shell(query_cmd) do |result|
+      expect(result.stdout).to match(%r{Last\sTUE})
```
michaeltlombardi: Should this be `TRUE` instead of `TUE`

cmccrisken-puppet: This is actually validating the response of `schtasks.exe /query /v /fo list /tn #{taskname}`:
```
Folder: \
HostName:                             ROMANTIC-FEMALE
TaskName:                             \mallard-duck
Next Run Time:                        2/25/2020 5:53:02 PM
Status:                               Ready
Logon Mode:                           Interactive only
Last Run Time:                        N/A
Last Result:                          1
Author:                               ROMANTIC-FEMALE\Administrator
Task To Run:                          C:\Windows\System32\calc.exe 
Start In:                             N/A
Comment:                              N/A
Scheduled Task State:                 Enabled
Idle Time:                            Disabled
Power Management:                     Stop On Battery Mode, No Start On Batteries
Run As User:                          ROMANTIC-FEMALE\Administrator
Delete Task If Not Rescheduled:       Enabled
Stop Task If Runs X Hours and X Mins: 72:00:00
Schedule:                             Scheduling data is not available in this format.
Schedule Type:                        Monthly
Start Time:                           N/A
Start Date:                           N/A
End Date:                             N/A
Days:                                 Last TUE
Months:                               Every month
Repeat: Every:                        Disabled
Repeat: Until: Time:                  Disabled
Repeat: Until: Duration:              Disabled
Repeat: Stop If Still Running:        Disabled
```
```
Days:                                 Last TUE
```

To be honest, the `idempotent_apply` of a manifest with `which_occurrence => last` verifies the functionality of this change, so I can drop that or, I can pick a different day! 😆

michaeltlombardi: I'm a derp, this is fine!

PR #123, opened 2020-02-21
replace puppet[:compatibility] with function call compatibility since…

fxbe:
Puppet resource command not working scheduled_task type and taskscheduler_api2 provider because validate uses @resource[:compatibility] variable which is not populated at the time the validation function is called. Replacing @resource[:compatibility] variable with call to compatibility method.

Error: Could not run: Validation of Scheduled_task[Microsoft\Configuration Manager\Configuration Manager Health Evaluation] failed: Microsoft\Configuration Manager\Configuration Manager Health Evaluation specifies a path including subfolders and a compatibility of 1 - tasks in subfolders are only supported on version 2 and later of the API. Specify a compatibility of 2 or higher or do not specify a subfolder path.

### DIFF
--- a/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
+++ b/lib/puppet/provider/scheduled_task/taskscheduler_api2.rb
@@ -208,7 +208,7 @@ Puppet::Type.type(:scheduled_task).provide(:taskscheduler_api2) do
   end
 
   def validate_name
-    return unless @resource[:name].match?(%r{\\}) && @resource[:compatibility] < 2
+    return unless @resource[:name].match?(%r{\\}) && compatibility < 2
 
     raise Puppet::ResourceError, "#{@resource[:name]} specifies a path including subfolders and a compatibility of #{@resource[:compatibility]}"\
                                  ' - tasks in subfolders are only supported on version 2 and later of the API. Specify a compatibility of 2 or higher or do not specify a subfolder path.'


michaeltlombardi:
Hi, @fxbe, thanks for this patch! Looking into it a little bit, it looks like you've actually stumbled across a bug, but your fix here will always default to a compatibility of `2` where the resource defaults to `1`.

You can verify that by running this manifest, which will not error on the validation block but _will_ fail when it tries to create the task:

```puppet
scheduled_task { 'test\example\Run Notepad':
  ensure   => present,
  command  => 'C:\Windows\System32\notepad.exe',
  trigger  => {
    schedule   => daily,
    start_time => '12:00',
  },
  provider => 'taskscheduler_api2'
}
```

If you add insert a pry just before https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/123/files#diff-2e1fd055fb04c2c02fbd4a6fe254184aR211 (`require 'pry' ; binding.pry`) and introspect on the values of `@resource[:compatibility]` and `compatibility` during the apply, you should see that they are `1` and `2` respectively.

Some investigation is needed to ensure that `@resource` is properly populated with the compatibility discovered for existing tasks without incorrectly defaulting when creating new tasks.

david22swan:
@fxbe Just checking in to see how your doing with this and whether you need any help?

david22swan:
@fxbe 
Making a quick change to point this PR towards the newly defined default main branch as part of our work to remove inappropriate terminology. Apologies if this is inconvenient in any way.

DavidS:
Hello, @fxbe !

It looks like this PR has gone over 30 days without any updates. Are you still interested in getting this merged? If so, please let us know and, if possible, address the concerns above. If you no longer have time or interest in this PR, that’s okay too, just let us know! If you are interested but need a little help, we’re happy to provide it!
We close stale PRs after 90 days of inactivity. If you don’t have time to get back to us right now, we’ll ping you again after the 60 day mark as a final reminder. Thanks!

david22swan:
@fxbe Hey, just jumping in to give you the 60 reminder.
If you need any help don't hesitate to shout :)

pmcmaw:
Hey @fxbe
 
I am going to close this PR due to inactivity. If you disagree with my decision or find time to work on this again please do not hesitate to reopen. I would also like to thank you for taking the time to submit this PR we really appreciate it. 👍

PR #145, opened 2020-07-21
(IAC-918) - `disable_time_zone_synchronization` function implemented

david22swan:


lib/puppet/type/scheduled_task.rb
```diff
@@ -140,6 +140,8 @@ def insync?(current)
           * `minutes_interval` --- The repeat interval in minutes.
           * `minutes_duration` --- The duration in minutes, needs to be greater than the
             minutes_interval.
+          * `disable_synchronise` --- Whether or not to disable the `synchronise across
```
sanfrancrisko, line 143: I would be tempted to name it something a bit more in line with the Windows functionality it's controlling. Perhaps `time_zone_synchronization`, or better still, exactly as it is in the UI: `synchronize_across_time_zones`. I will admit that's a bit long, but it will make it obvious what it's related to.

spec/acceptance/should_create_task_spec.rb
```diff
@@ -251,4 +251,56 @@
...
+        start_time          => '12:00',
+        disable_synchronise => true,
+      },
+      provider    => 'taskscheduler_api2'
+    }
+    MANIFEST
+    apply_manifest(pp, catch_failures: true)
+
+    # Ensure it's idempotent
+    # apply_manifest(pp, catch_changes: true)
```
sanfrancrisko, line 274: Can this be deleted?

spec/acceptance/should_create_task_spec.rb
```diff
@@ -251,4 +251,56 @@
...
+        start_time          => '12:00',
+        disable_synchronise => true,
+      },
+      provider      => 'win32_taskscheduler'
+    }
+    MANIFEST
+    apply_manifest(pp, catch_failures: true)
+
+    # Ensure it's idempotent
+    # apply_manifest(pp, catch_changes: true)
```
sanfrancrisko, line 300: Can this be deleted?

spec/acceptance/should_create_task_spec.rb
```diff
@@ -251,4 +251,56 @@
...
+
+    # Verify the task exists
+    query_cmd = "schtasks.exe /query /v /fo list /tn #{taskname}"
+    run_shell(query_cmd)
```
sanfrancrisko, line 278: I know this is how we verify using existing tests, but if the output that comes back from this query gives us any indicator that the `Synchronize across time zones` parameter is `on|off`, `true|false`, `0|1`, etc... it would be good to verify that

david22swan: It doesn't. I checked and it always gives the local time regardless if it's UTC or not.

spec/acceptance/should_create_task_spec.rb
david22swan: Removed and added a note to the type to show that it is non-idempotent

spec/acceptance/should_create_task_spec.rb
david22swan: Removed and added a note to the type to show that it is non-idempotent

spec/acceptance/should_create_task_spec.rb
sanfrancrisko: No worries - just thought I'd check.

lib/puppet/type/scheduled_task.rb
david22swan: Changed to `disable_time_zone_synchronization`. `synchronize_across_time_zones` doesn't really work due to the function turning it of rather than varying it or turning it on.

PR #150, opened 2020-07-29
(IAC-732) - implement `Run only when user is logged on

david22swan:


REFERENCE.md
```diff
@@ -66,6 +66,10 @@ Please also note that Puppet must be running as a privileged user
 in order to manage `scheduled_task` resources. Running as an
 unprivileged user will result in 'access denied' errors.
 
+If a user is specified without an accompanying password the
```
sanfrancrisko, line 69:
```suggestion
If a user is specified without an accompanying password, the
```

david22swan: fixed

PR #158, opened 2020-08-24
(MODULES-10783) Add back empty? check for `datetime_string` value

sanfrancrisko:
In https://github.com/puppetlabs/puppetlabs-scheduled_task/pull/145 it
looks as though we removed the empty check on `datetime_string`.

This was causing the PE Patch Management tasks to fail in PE 2019.8.1.


sanfrancrisko:
**Release checks:**

image

PR #161, opened 2020-08-25
(bugfix) - fix `disable_time_zone_synchronization` so that it correctly disables functionality when set to true

david22swan:


sanfrancrisko (commented):
I'm wondering if we should take a look at what the test coverage here given we've got 2 fixes going in to this functionality in as many days?

Would be happy for this to be merged for now, to get the fix out there, but I think we need to look at the test coverage. 

If we think enhancing the tests is quick, then let's do it, otherwise we can merge and raise a ticket to address after.

sanfrancrisko (approved):

PR #175, opened 2020-12-16
(MODULES-10893) Fix Last Day Of Month Trigger

RandomNoun7:
Prior to this change, attempting to create a trigger that will run on
the last day of each selected month will fail. The error claims that a
a value is out of range

```
OLE error code:0 in <Unknown>
      <No Description>
    HRESULT error code:0x8002000a
      Out of present range.
```

Microsoft [documentation] claims in one place that the correct way to
specify the last day of a month is to effectively translate a bitmask
value of day 32 and pass that into the `{get:set}_DaysOfMonth` method
on the trigger object. The module was previously coded to do exactly
this. The provider and it's helpers would catch the string value 'last'
in the `on` property of a `monthly` trigger and translate it into the
bit mask. This is what resulted in the error, because it turns out that
the position for day 32 in the bit mast is outside the acceptable range
of values for this property.

This PR changes strategy and instead relies on a different section
of the documentation in which an entirely different property on the
trigger object called [RunOnLastDayOfMonth] is set.

The code now catches the string value 'last' in the `on` property of
the trigger, and sets this boolean on the trigger object.

Testing has shown this to be an effective method of managing this
property of a trigger.

[documentation]: https://docs.microsoft.com/en-gb/windows/win32/api/taskschd/nf-taskschd-imonthlytrigger-get_daysofmonth
[RunOnLastDayOfMonth]: https://docs.microsoft.com/en-gb/windows/win32/api/taskschd/nf-taskschd-imonthlytrigger-put_runonlastdayofmonth

### DIFF
--- a/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
+++ b/lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
@@ -5,7 +5,8 @@ require 'time'
 # @api private
 # PuppetX::PuppetLabs::ScheduledTask module
 module PuppetX::PuppetLabs::ScheduledTask
-  # This module is used to manage the triggers for the Task Scheduler V2 API
+  # @api private
+  # PuppetX::PuppetLabs::ScheduledTask::Trigger module
   module Trigger
     # Gets or sets the amount of time that is allowed to complete the task.
     class Duration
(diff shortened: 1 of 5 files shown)


lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
```diff
@@ -389,42 +390,44 @@ def self.bitmask_to_names(bitmask)
...
         def self.indexes_to_bitmask(day_indexes)
-          day_indexes = [day_indexes].flatten.map do |m|
-            # The special "day" of 'last' is represented by day "number"
-            # 32. 'last' has the special meaning of "the last day of the
-            # month", no matter how many days there are in the month.
-            # raises if unable to convert
-            (m.is_a?(String) && m.casecmp('last').zero?) ? 32 : Integer(m)
+          integer_days = [day_indexes].flatten.select { |i| i.is_a?(Integer) }
+          invalid_days = integer_days.reject { |i| i.between?(1, 31) }
+          if day_indexes.nil? || (day_indexes.is_a?(Hash) && day_indexes.empty?)
```
sanfrancrisko, line 410: Should we move this check up to the start of the method on line 408, before we do anything else with `day_indexes`?

sanfrancrisko (changes requested):
Just one minor issue, but overall looks good 👍 

Thanks for picking this up and addressing - took some work off my plate today 😃

lib/puppet_x/puppetlabs/scheduled_task/trigger.rb
RandomNoun7: Yup. We can do that.

.rubocop.yml
```diff
@@ -47,6 +47,7 @@ Style/BracesAroundHashParameters:
 Style/ClassAndModuleChildren:
   Description: Compact style reduces the required amount of indentation.
   EnforcedStyle: compact
+  Enabled: false
```
RandomNoun7, line 50: This change is required because of the module declaration changes that I will comment on in that file.

sanfrancrisko (approved):
Excellent, thanks for this @RandomNoun7 👍

DavidS (approved):

PR #207, opened 2022-03-31
Make disable_time_zone_synchronization idempotent

omolenkamp:


chelnak:
Hey @omolenkamp, thanks for the contribution!

Please could you provide a little more context around this change?

Also it would be great if you could add a test or two for the new functionality.

Let us know if you need any help.

omolenkamp:
`disable_time_zone_synchronization` works by setting `StartBoundary` to a local timestamp (no zone offset) when `true`, and using a timestamp including a zone offset when `false`. Because of this, the current value of `disable_time_zone_synchronization` can be determined by testing if `StartBoundary` includes a zone offset. (`Z`, `+xx:xx`, or `-xx:xx`)

(`start_boundary` cannot be used here, as the Ruby `Time` class doesn't differentiate between local and zoned timestamps)

PR #240, opened 2023-09-25
(MAINT) - Fixing rubocop failures

rajat-puppet:
## Summary
Fixing rubocop failures

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -160,6 +160,11 @@ RSpec/NoExpectationExample:
     - 'spec/unit/puppet/provider/scheduled_task/win32_taskscheduler_spec.rb'
     - 'spec/unit/puppet/type/scheduled_task_spec.rb'
 
+# Offense count: 1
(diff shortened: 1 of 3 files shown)


Ramesh7 (changes requested):
can we fix it over putting into skipping list?

rajat-puppet:
> can we fix it over putting into skipping list?

The issue is not with the file

> can we fix it over putting into skipping list?

The issue is not with the file but with the way describe block has been written to use Trigger model using the namespace, tried correcting it but didn't make any sense, so I put it up on the exclude list.

PR #251, opened 2024-04-16
Fix spelling of which_occurrence in REFERENCE.md

rychannel:
## Summary
'which_occurrence' parameter is spelling 3 different ways in REFERENCE.md. This fixes that.

## Additional Context
Team member couldn't figure out why code didn't work after copying example.


jst-cyr (approved):
This is text-only fix for spelling which is still wrong in the main branch after a rebase. Tests are failing, but are unrelated to this change.

